PR #87, opened 2015-06-27
significant performance boost in Kore VertexBuffer

JeriXNet:
I've discovered this fix with MSVS 2013 Profiler and this tool still point to `memcpy` in this fix as one of the functions doing most individual work:

This fix requires numerous edits in Kore VertexBuffer backends, something like:
```
    public:
        int myCount;
        int myStride;
    protected:
```

### DIFF
--- a/Backends/Kore/kha/graphics4/VertexBuffer.hx
+++ b/Backends/Kore/kha/graphics4/VertexBuffer.hx
@@ -9,6 +9,7 @@ import kha.graphics4.VertexStructure;
 @:headerCode('
 #include <Kore/pch.h>
 #include <Kore/Graphics/Graphics.h>
+#include <string.h>
 ')
 
 @:headerClassCode("Kore::VertexBuffer* buffer;")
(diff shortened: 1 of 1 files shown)


RobDangerous:
Actually planning to get rid of that data copy completely by changing Float32Array (or creating a special version of Float32Array for Kha).
But let's use this for now, until I do that. Did using myCount/myStride directly instead of stride()/count() speed things up? Because at least in theory Visual Studio's whole program optimization should get rid of that.

JeriXNet:
Yes, direct usage of myCount/myStride really increased speed that's confused me to.

RobDangerous:
In that case it makes sense - debug mode generally doesn't inline anything, especially not across compilation units.
But anyway, I just applied the optimization I had originally in mind. Can you test your changes against the update I just checked in?

JeriXNet:
Yeah! I've just tested it and it works great!
Therefore this pull request should be rejected.

RobDangerous:
Okidoki :+1:

PR #103, opened 2015-07-09
make sure the canvas size in pixels match the canvas size on screen

wighawag:
This small change ensure the browser do not scale the content of the canvas to fit the visual size of the canvas. 
While this has no benefit when used with a fixed size canvas but benefit all the other cases.

### DIFF
--- a/Backends/HTML5/kha/Starter.hx
+++ b/Backends/HTML5/kha/Starter.hx
@@ -189,6 +189,21 @@ class Starter {
 			Scheduler.executeFrame();
 			
 			if (canvas.getContext) {
+
+				// Lookup the size the browser is displaying the canvas.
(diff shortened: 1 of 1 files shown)


ratalaika:
I recommend you to correct the tabs to adjust to original Kha tab.

Btw for what kind of cases this would be used? So I can document it on the wiki :)

wighawag:
Basically, now if the canvas is resized it will still be of the correct number of pixels (to match the new size)
Currently the html template for the index.html make the canvas no resizable so it is not affected but if the template make the canvas responsive to window size (which it need to be for mobile), it will alway have the same number of pixels as its display size.

wighawag:
fixed the ident by the way

RobDangerous:
Nicely done, thanks.

PR #128, opened 2015-08-07
Multitouch support for Html5

RafaelOliveira:
This brings back setTouchXY in touchUp, and add support for multitouch. Based on this article:
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events


RobDangerous:
Awesome, thank you very much!

PR #129, opened 2015-08-11
Implemented integer arithmetic for Vector2i.

AoE-Maniac:
Ported almost all Vector2 functions to Vector2i (left out normalization because it doesn't make sense with integers).


RobDangerous:
Thank you very much!

PR #132, opened 2015-08-13
Added depth buffer for rendering to textures in html5.

AoE-Maniac:
Tested with firefox and chrome.


RobDangerous:
Looking good.

PR #135, opened 2015-08-18
AntiAliasing support for OSX

luboslenco:


RobDangerous:
Build failed - but it only did because the Kore reference wasn't updated, so I'll take this anyway.
Thanks!

PR #157, opened 2015-09-13
fix html5 video size

wighawag:


RobDangerous:
Very nice, thank you.

PR #165, opened 2015-09-23
More work on scheduler: Reused delete lists and removed two unused va…

AoE-Maniac:
…riables.

As discussed, no fancy new features or changes (yet?).


RobDangerous:
Thank you very much.

PR #176, opened 2015-10-11
add mouse lock for Kore

wighawag:


RobDangerous:
Code looks great, thank you very much.

PR #177, opened 2015-10-15
Clean up log

luboslenco:
Hope its ok, as it can get annoying to have too much stuff in console.


wighawag:
Sorry about that, should not be there. Was for testing only thanks for fixing it

luboslenco:
No prob, I do that all the time:)

PR #180, opened 2015-10-22
add fullscreen switch calls for html5 + Kore + empty implementations …

wighawag:
…for the rest

The Kore implementation has been tested only on windows and because the changeResolution function on Windows is not setting the window to fullscreen, it does not work yet. 
But the call  from haxe should stay like this.


wighawag:
this pull request also include a small fix to MouseLock so that you cannot lock the mouse twice and get fall mouselock change callback (in Kore kha/Starter.hx)

RobDangerous:
Thanks a lot!

PR #187, opened 2015-11-13
In HTML5 target : Added mouse wheel event detection, added middle mou…

memilian:
…se button detection in mouse click events.
should fix https://github.com/KTXSoftware/Kha/issues/182. Other targets does not seems to all detect middle click though


RobDangerous:
Thanks!

PR #194, opened 2015-12-16
add Khanvas width/height setting on init and changeResolution

laxa88:
HTML5 target only for now

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -44,6 +44,9 @@ class SystemImpl {
 	
 	public static function init(title: String, width: Int, height: Int, callback: Void -> Void) {
 		init2();
+
+		changeResolution(width, height);
+
 		//MouseImpl.init();
 		callback();
 	}
@@ -722,6 +725,7 @@ class SystemImpl {
 	}
 
 	public static function changeResolution(width: Int, height: Int): Void {
-		
+		khanvas.width = width;
+		khanvas.height = height;
 	}
 }


wighawag:
This is not going to work properly if at all?,
setting canvas.width/height set the number of pixels for that canvas buffer not its actual size.
Also the width/height is set every frame to make sure to map one to one to the current display size of the canvas (which is accessble via the read only clientWidth/clientHeight)
You can use css style to redimension your canvas and kha as stated in the previous sentence will automatically adjust the frame size.
I think the best way to consider the html5 target is like a mobile where you cannot change its resolution. The resolution being set in the html.

laxa88:
the problem is, the width/height on khanvas/canvas is always 800x600 no matter how you change it. you can try it out yourself.

wighawag:
Did you try changing the index.html in build/html5 ?

laxa88:
yes i did. my point is, setting width/height in System.init() is pointless if I need to go into the index.html to change the canvas size manually. The function is empty anyway, so wouldn't it be convenient to have it in for HTML5 backend? :D

wighawag:
I think it is preferable to think as the canvas as a non changeable from the game point of view (like a mobile screen). In this case, this is the responsability of the page to setup the canvas size (hence css)

What your particular use case?

RobDangerous:
The parameters in init are just hints which only work for desktop applications. The nice thing about having to do that in the html file is that it makes it easier to conveniently embed the canvas in an actual website in any way you want.

laxa88:
i want to be able to change the screen resolution at runtime depending on the size of the browser. i.e. on mobile phones, it would be 320 x 480. If it's on the desktop browser, it becomes 768 x 1366. but okay. if this is unnecessary for kha then i'll just manually do it from wyngine.

wighawag:
you can do that with css (lookup media query)

laxa88:
I'm not sure if we're talking about the same thing here. The CSS only affects the style of the khanvas. I am looking to change the actual resolution of the khanvas. Since the default kha game size is 800x600 (with no CSS), I will have to manually edit the html to make it start from 320 x 480. Even when using media query e.g.

    @media (max-width:600px) {
        #khanvas {
            width: 100px;
            height: 100px;
        }
    }

only the size of the khanvas changes (e.g. to 100x100), but the resolution remains 320 x 480. If I wanted the resolution to change to 100x100 as well, I would have to modify the khanvas's "width" and "height" property, which I did in this pull request. Is there any other way to do it?

laxa88:
Oh I think I just embarrassed myself. Just experimented with it more, and realised @media actually modifies the khanvas html's width/height rather than the CSS width/height. I'm gonna play around with that and see if it works out for me. Thanks @wighawag !

ashes999:
Even though this was closed, I would like to add my voice to this: I think it's a bug. I specified my game size in `System.init` as 576x1024, and I get an 800x600 app. I can't keep manually changing the index.html file or any of the generated files every single time I build.

During development, is there a way to make this work the way I expect?

ashes999:
There's also [this PR](https://github.com/Kode/khamake/pull/30) which allows you to set `windowOptions.width` and `windowOptions.height` in `khafile.js`. But this doesn't work if you're using Kha Kha `16.1.2` from haxelib.

RobDangerous:
Please don't use haxelib Kha, it's deprecated.

ashes999:
@RobDangerous I don't like using Git unless there are discrete (stable?) versions that I can use. It looks like the Kha git repo doesn't have tags. That makes it more difficult to get the same version of Kha across multiple machines (I have to write down the commit hash somewhere).

Am I missing something, or is there a better way to manage this?

RobDangerous:
Hm, I can add tags in the future.

ashes999:
It looks like it takes several minutes to clone Kha and init all the subrepositories. It also looks like I have to take extra steps to make it accessible to the command-line.

Is there a reason you don't support publishing to haxelib any more? It really simplifies deployment and getting to a consistent and/or stable version.

sh-dave:
Just add Kha as a submodule to your project, then you don't need tags and whatnot.

ashes999:
@RobDangerous I would very much appreciate it if you could add tags. Thanks.

RobDangerous:
Will do. Just started by adding a tag for the version included in the latest Kode Studio release.

PR #195, opened 2015-12-16
setFloats() array size

luboslenco:
Hello,

Any reason to max this out at 100?

I finally got to implement GPU skinning, and did so using setFloats(). For 50 bones I need array of 600 floats(12 per bone). Everything seems to work fine, but I'm wondering if there would be any difference in speed using float[600] vs vec4[150] (non-existant on Kha).


RobDangerous:
Putting a dynamic array on the stack like this only works with some compilers. Let's figure out how to do that properly tomorrow.

RobDangerous:
Fixed that, please test if it works for you.
float[600] vs vec4[150] shouldn't make any difference at all (in gpus basically everything is a vec4 anyway).

PR #200, opened 2016-01-02
HTML5 Canvas fix clear(), add drawCircle/fillCircle handle for Canvas…

laxa88:
… in GraphicsExtension

Sorry for the long message!


RobDangerous:
Looking good, thank you.

PR #218, opened 2016-01-23
Fix application state callbacks

laxa88:
Tested on android-native and HTML5 with code below:

    System.notifyOnApplicationState(function () {
			// foreground
			trace("onForeground");
		}, function () {
			// resume
			trace("onResume");
		}, function () {
			// pause
			trace("onPause");
		}, function () {
			// background
			trace("onBackground");
		}, function () {
			// shutdown
			trace("onShutdown");
		});

In HTML5 target, only onBackground (onblur) and onForeground (onfocus) is triggered.


RobDangerous:
Thank you very much!

PR #226, opened 2016-01-30
Made transformation between glViewport and glScissor consistent.

AoE-Maniac:
In glScissor y inversion ist taken into account, but not in glViewport. This results in incompatible behaviour when called with the same values. I therefore suggest also making this transformation in glViewport.

### DIFF
--- a/Backends/HTML5/kha/js/graphics4/Graphics.hx
+++ b/Backends/HTML5/kha/js/graphics4/Graphics.hx
@@ -99,7 +99,8 @@ class Graphics implements kha.graphics4.Graphics {
 	}
 
 	public function viewport(x: Int, y: Int, width: Int, height: Int): Void{
-		SystemImpl.gl.viewport(x,y,width,height);
+		var h: Int = renderTarget == null ? System.pixelHeight : renderTarget.height;
+		SystemImpl.gl.viewport(x, h - y - height, width, height);
 	}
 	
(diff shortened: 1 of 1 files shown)


wighawag:
Why do we need to do that? Is it the reverse on non html5 targets ?

AoE-Maniac:
Yes, y is reversed on some targets (like WebGL or OpenGL) but not on others (like Direct3D). If this transformation would not be done in the backend you would have to check renderTargetsInvertedY() each time a viewport or scissor is created and then change the parameters accordingly. So doing that in the backend saves lots of trouble and is more intuitive when targeting multiple platforms.
Plus the transformation was alread done in scissor, which made it even more confusing when using the functions in tandem.

RobDangerous:
Plus renderTargetsInvertedY will soon disappear and be handled by the shader compiler automatically.

AoE-Maniac:
Wohoo, had something like than in mind too - but then thought "Na, seems lots of work and Robert has to much on his list already." :-D

RobDangerous:
Weirdly enough that'll get me into some trouble with only one target: http://docs.unity3d.com/Manual/SL-PlatformDifferences.html

AoE-Maniac:
So you are saying that unity can never be achieved as long as there is unity?

RobDangerous:
I'll unite everything by compiling all shaders twice.

PR #231, opened 2016-02-02
Added stencil bindings for kore.

AoE-Maniac:
Succesfulls tested with Kharira on Windows using OpenGL.

Enum conversion feels a bit hacky, maybe there is some room for improvement there.


RobDangerous:
Enum conversion is as good as it gets.

PR #232, opened 2016-02-03
Enabled depth + stencil buffer in flash target

sh-dave:
Fixed wrong parameter given to setStencilParameters (stencilReadMask instead of stencilReferenceValue)

Not sure if it's a good idea to hardcode context.configureBackBuffer(..., true), maybe add some SystemOptions to kha.System.init()? I kinda like the luxe style of passing options, so you don't end up with a gazillion parameters. Like

```haxe
kha.System.init({
    title : 'YadaYada',
    width : 800, height : 600,
    depthAndStencil : true,
}, myCallback);
```

### DIFF
--- a/Backends/Flash/kha/SystemImpl.hx
+++ b/Backends/Flash/kha/SystemImpl.hx
@@ -60,7 +60,7 @@ class SystemImpl {
 	
 	private static function onReady(_): Void {
 		context = stage3D.context3D;
-		context.configureBackBuffer(width, height, 0, false);
+		context.configureBackBuffer(width, height, 0, true);
 		keyboard = new Keyboard();
 		mouse = new MouseImpl();
 
@@ -244,7 +244,7 @@ class SystemImpl {
 	
 	private static function resizeHandler(event: Event): Void {
 		if (frame != null && stage.stageWidth >= 32 && stage.stageHeight >= 32) {
-			context.configureBackBuffer(stage.stageWidth, stage.stageHeight, 0, false);
+			context.configureBackBuffer(stage.stageWidth, stage.stageHeight, 0, true);
 		}
 	}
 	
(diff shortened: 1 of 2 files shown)


RobDangerous:
Thanks.
And I agree, we can pass an options object for init - but only for init, because anonymous objects are very slow when cross-compiled to C++.

wighawag:
if the function is inlined the anonymous object should be inlined as well, no (https://github.com/HaxeFoundation/haxe/issues/4172)?
I wish haxe had an option to enforce the use of anonymous object so we know it will be inlined from the function point of view. (as if you pass an non anonymous object it will not be inlined) This would allow to pass argument by name without cost

sh-dave:
The how doesn't really matter to me. Maybe a builder pattern (iirc) works even better, b/c other than typedefs

1) you get code completion (at least in FlashDevlop it doesn't really work that well for me)
2) you can see the defaults in code completion already

```haxe
class SystemOptions {
    public function new( title : String, width : Int = 800, height : Int = 600 ) {
     ....
    }

    public function setDepthAndStencil( value : Bool ) : SystemOptions {
       this.depthAndStencil = value;
       return this;
    }
}


kha.System.init(new SystemOptions('YadaYada', 512, 512)
    .setDepthAndStencil(true)
    .setOtherOptionalStuff()
);
```

RobDangerous:
I like the first option better because it's simpler. Also I just implemented that.

sh-dave:
Ok cool, i'm in the process of implementing multiple window support (Win/Linux with OpenGL), needed this as well and went with the 2nd option. But i will change my stuff to the typedef style.

RobDangerous:
That's excellent. Oh and I want to formally invite you to join the Kha irc chat any time.

PR #233, opened 2016-02-03
Updated readme

luboslenco:
I mentioned idea of making the readme shorter and removing content that is duplicated from wiki(and link to wiki instead). Let me know what do you think.


RobDangerous:
I like it.

PR #237, opened 2016-02-08
Android/HTML5: Support for depth/stencil buffers

sh-dave:
Hi Robert,

this PR adds support for depth/stencil buffers on android.
I'm not quite sure it's the best way to implement it. I tried using a separate attached stencil/depth buffer, but that fails on the s4 and succeeded but didn't work on my htc. Now i use the GLES1.1 GL_DEPTH_STENCIL_OES extension and it's working on all devices i have access to so far (HTC Desire X, S3, S4, HUAWEI Ascend something).

Edit: Oh it seems that the 'HTML5 stencil support' commit got somehow added as well (i'm still very much a noob in using git; I thought a pull request is just for a specific commit. I wanted to create a separate PR for it so they stay small and easier to merge).

### DIFF
--- a/Backends/Android/android/opengl/GLES11Ext.hx
+++ b/Backends/Android/android/opengl/GLES11Ext.hx
@@ -1,0 +1,7 @@
+package android.opengl;
+
+class GLES11Ext {
+	public static var GL_DEPTH_STENCIL_OES: Int;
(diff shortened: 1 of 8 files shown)


RobDangerous:
Thanks for testing on all of those devices, couldn't even have done that.

PR #246, opened 2016-02-18
Initial audio1 implementation for Unity backend.

magneticrealms:
Here's my initial audio1 implementation for Unity.

I wasn't sure how to finish off getting MusicChannel working so I've left it commented out in Backends/Unity/kha/audio1/Audio.hx for now.

On line 48 of Backends/Unity/kha/audio1/UnitySoundChannel.hx I've explicitly written the C# code including a typecast for set_volume() to avoid a compiler error about implicit conversion from double to float - is there a better way to do this from the haxe side?

In Sources/kha/audio1/AudioChannel.hx I've changed some private functions to public - I assume that's how they are supposed to be defined?

Also I'm not sure about loading the sound in the UnitySoundChannel constructor - is there a better place I could put that? I'm not totally clear on how asset loading works in Kha yet.

### DIFF
--- a/Backends/Unity/UnityBackend.hx
+++ b/Backends/Unity/UnityBackend.hx
@@ -2,10 +2,12 @@ package;
 
 import haxe.io.BytesData;
 import unityEngine.Texture2D;
+import unityEngine.AudioClip;
 
 extern class UnityBackend {
 	public static function uvStartsAtTop(): Bool;
 	public static function loadImage(filename: String): Texture2D; 
 	public static function loadBlob(filename: String): BytesData;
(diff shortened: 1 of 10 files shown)


RobDangerous:
The AudioChannel functions should stay private. You don't call them directly, they are just here so you can write channel.length, see http://haxe.org/manual/class-field-property.html
Can you add another commit to fix that?
Don't bother about the MusicChannel, it's deprecated.
Everything else looks good.

magneticrealms:
Sorry about that, not sure what I was thinking there!

RobDangerous:
Don't worry, I'm very glad you're helping out.

PR #252, opened 2016-02-24
Unity audio1 improvements.

magneticrealms:
Made the Unity audio1 implementation more robust and added code to clean up AudioSource components when they are no longer needed.


RobDangerous:
Thanks!

PR #257, opened 2016-03-03
Changes in KhaActivity to handle onActivityResult

RafaelOliveira:
I'm creating a bluetooth extension for android-java, and I need to call KhaActivity.startActivityForResult (to ask the user to enable the bluetooth), and receive the result in KhaActivity.onActivityResult.
So I created a list of "ActivityResults" that are extensions that are registered to receive the results from onActivityResult.
What do you think.

### DIFF
--- a/Backends/Android/android/app/Activity.hx
+++ b/Backends/Android/android/app/Activity.hx
@@ -1,6 +1,7 @@
 package android.app;
 
 import android.content.Context;
+import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Bundle;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


RafaelOliveira:
This is for android-java.

RafaelOliveira:
I closed because I want to test a better idea. I have some code that need to be called on onDestroy. So would be better a Extension class, with methods like onPause, onResume, onDestroy and onActivityResult. This class would be extended by the extension, and the methods would be called in the KhaActivity.

RobDangerous:
Oh, okidoki :)

PR #258, opened 2016-03-03
New class KhaExtension to handle extensions life cycle

RafaelOliveira:
I created the class KhaExtension that can be registered to be called in some events in the KhaActivity. In my case I need this to use the events onDestroy and onActivityResult of the activity.

Basically a extension extends KhaExtension, register itself with KhaActivity.the().registerExtension(), and then override the functions that will be called in the KhaActivity.

Check if the code is good.

### DIFF
--- a/Backends/Android/android/app/Activity.hx
+++ b/Backends/Android/android/app/Activity.hx
@@ -1,6 +1,7 @@
 package android.app;
 
 import android.content.Context;
+import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Bundle;
 import android.view.View;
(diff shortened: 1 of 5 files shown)


RobDangerous:
Actually this is great. Thanks a lot!

RafaelOliveira:
I remembered that Kha has functions for pause, resume, and shutdown events, using System.notifyOnApplicationState. I could use that, so I think that maybe my code is redundant, the only problem is that after a test the shutdown event isn't being called, but, at least is there. The only part that is specific to android is the onActivityResult method. So I was thinking if I should send another pr with just the onActivityResult part. What do you think.

RobDangerous:
Yes, you should be able to use that. Sorry, I was too fast in accepting the PR. Also please note the queueEvent calls which are used to push the events to the correct thread (Android apis are run on a bunch of different threads and care has to be taken to run stuff on the proper threads).

PR #259, opened 2016-03-05
Fix to get the correct size of the screen in System.windowWidth/Heigh…

RafaelOliveira:
…t in android-java


RobDangerous:
Cool. We should probably cache the DisplayMetrics object though.

PR #266, opened 2016-03-15
renamed audio2.Audio.playMusic() to audio2.Audio.play()

sh-dave:
I'm trying this and couldn't compile windows target b/c kha.audio2.Audio.play wasn't found (It does compile on flash).

```haxe
var channel : kha.audio1.AudioChannel = kha.audio2.Audio.play(sound, loop);

if (channel == null) {
    channel = kha.audio1.Audio.play(sound, loop);
}
```

Was this just overlooked when you renamed stuff or am i using it wrong?


RobDangerous:
Just an oversight.

PR #281, opened 2016-04-04
Edit "Get Started" URL.

hazagames:
[http://kha.tech/download/](http://kha.tech/download/) to [http://kha.tech/download](http://kha.tech/download). [http://kha.tech/download/](http://kha.tech/download/) returns 404.


RobDangerous:
Thanks!

PR #289, opened 2016-04-12
requestContext3D expects a String

juakob:


RobDangerous:
Due to the Haxe update. Thanks!

PR #291, opened 2016-04-14
Made charcode sending language independent

jefvel:
I have a öäå layout and noticed I got some other characters instead. I'm not sure what would be a good way to detect what key is down, since it will return "1" if no shift is pressed, and "!" if shift is pressed, and so on.


jefvel:
Also fixed scrolling in html5

PR #292, opened 2016-04-15
Fixed mouse wheel in html5 target

jefvel:


memilian:
This removed the wheel support for firefox. FF does not have any mousewheel event, just a wheel event

RobDangerous:
Looks like we're running in circles but that kind of fits the topic.

jefvel:
hehe, yeah. Assigning both onwheel and onmousewheel seemed to fix it. However the deltaY values differ in different browsers.

jefvel:
#293 there's a pull request. is it possible to reopen pull requests?

RobDangerous:
Maybe. If not you can just create a new one from the same branch.

PR #293, opened 2016-04-18
Fixscroll

jefvel:
mousewheel in firefox and IE.


jefvel:
the deltaY values differ depending on the browser used. Should this be normalized in the kha backend?

RobDangerous:
Ideally yes.

jefvel:
I made it so it returns deltaY in the amount of clicks in the mousewheel. It works the same in chrome, IE and firefox.

PR #295, opened 2016-04-22
Fix mouse up issue on HTML5 target caused by releasing a button while another is being held.

hazagames:
Currently the mouse up listener will be removed each time the event gets dispatched so if multiple buttons are held at the same time,  mouseUp will get called only for the first released button. This commit fixes that problem.


RobDangerous:
Excellent!

PR #305, opened 2016-05-01
Added Display.widthPrimary and Display.heightPrimary

RafaelOliveira:
Display.widthPrimary and Display.heightPrimary returns the size of the primary display. I just used the code that was already in Kore. The only version implemented is windows, there is code to get the primary display in linux but is crashing in my system, so I removed the implementation. The Pi was following the linux implementation, but I don't have a Pi to test for now so I removed too.

### DIFF
--- a/Backends/Android/kha/DisplayImpl.hx
+++ b/Backends/Android/kha/DisplayImpl.hx
@@ -14,4 +14,14 @@ class DisplayImpl {
         trace('TODO (DK) implement me');
         return -1;
 	}
+
+	public static function widthPrimary() : Int {
+        trace('TODO (DK) implement me');
+        return -1;
+	}
+
+	public static function heightPrimary() : Int {
+        trace('TODO (DK) implement me');
+        return -1;
+	}
 }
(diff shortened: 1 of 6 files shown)


luboslenco:
Is there a way to get id of primary display? That could be handy on it's own and also used in Display.width(primaryDisplayId).

RafaelOliveira:
In windows the primary display is checked here
https://github.com/KTXSoftware/Kore/blob/master/Backends/Windows/Sources/Kore/Display.cpp#L82
but only the info is stored, not the id. I can remove the Display.widthPrimary/heightPrimary and put something like getPrimaryDisplayId().

RafaelOliveira:
Actually, there is more info available in the Display class (at least in Windows and Linux), like the x and y position of the display, that could be used for things like passing a object from one display to another in the same position. I think all the info could be exposed.
@sh-dave I'm correctly about the x/y positions of the display? Because we can arrange the displays in any position we want in windows (in relation to each other).
I will just implement getPrimaryDisplayId() and leave this idea for later, because the implementation is more complicated.

sh-dave:
I haven't found a really reliable way to identify the display's on linux (using libxinerama). Primary seems to always be the first display returned though. But i actually wanted to get the Id's that are displayed when you click on 'Identify' on windows, or whatever id is shown in the display configuration tool on linux. There is another library to get the info libxrandr, but i haven't had time to look into it yet.

And yes, the x/y is the displays position, i think they are always relative to the primary (which is at [0; 0])

PR #307, opened 2016-05-04
Add worldSpace option for drawing polygons

Devination:
Found myself wanting a worldSpace option when drawing polygons. Not sure if this is the best way to do it or if it fits your code style though

### DIFF
--- a/Sources/kha/graphics2/GraphicsExtension.hx
+++ b/Sources/kha/graphics2/GraphicsExtension.hx
@@ -83,32 +83,44 @@ class GraphicsExtension {
 	/**
 	 * Draws a convex polygon.
 	 */
-	public static function drawPolygon(g2: Graphics, x: Float, y: Float, vertices: Array<Vector2>, strength: Float = 1) {
+	public static function drawPolygon(g2: Graphics, x: Float, y: Float, vertices: Array<Vector2>, strength: Float = 1, ?worldSpace: Bool = false) {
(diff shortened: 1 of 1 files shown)


RobDangerous:
Doesn't make too much sense to me, you could just pass in 0 for x and y, right?

Devination:
Not quite, if you look at the last 2 arguments in the fillTriangle calls, x and y are also used for the last point of every triangle in the polygon. So while 0s for x and y will kill the offset, it will also pin 1 point (the 'center' point) of the polygon to the top left corner of the screen. Wednesday, 04 May 2016, 08:28AM -07:00 from Robert Konrad < notifications@github.com> :

RobDangerous:
Uh, that is very strange. Any objections if I change that? Don't think a center point makes sense.

Devination:
No objections from me, sounds like a good change! I'm just not sure how extensively other people are relying on it in it's current form.

Devination:
Or wait sorry, (x,y) doesn't have to be a center point. It could just as well be a corner of the polygon from which the rest of the verts fan out. It's a little strange that you supply those separately from the rest of the verts instead of just taking them from the first vert in the array but that doesn't bother me as much as not having a world space option

RobDangerous:
It's a so called triangle fan. But I think having a regular triangle array instead makes more sense and resolves the world space problem.

RobDangerous:
Hey, there are triangles nowadays.

PR #312, opened 2016-05-10
Image.fromBytes with WebGL implementation.

luboslenco:
Need a fast way to construct images from already allocated bytes, this is the most efficient/straightforward way I figured. Thoughts?


RobDangerous:
Looks good.

PR #334, opened 2016-07-04
lockMouse didn't work in firefox. it does now

jefvel:


RobDangerous:
Oh, this would also work without getElementById by just using SystemImpl.khanvas instead of canvas, right?

Backends/HTML5/kha/SystemImpl.hx
```diff
@@ -386,15 +386,19 @@ class SystemImpl {
 	}
 
 	public static function lockMouse(): Void {
+        var canvas: Dynamic = Browser.document.getElementById("khanvas");
+
 		untyped if (SystemImpl.khanvas.requestPointerLock) {
 			SystemImpl.khanvas.requestPointerLock();
-		}
-		else if (canvas.mozRequestPointerLock) {
```
wighawag: the simplest change would be to replace "canvas" with SystemImpl.khanvas on  that line, no ?

RobDangerous:
Implemented the simpler fix, mentioned you in the commit.

PR #337, opened 2016-07-07
 fast matrix and blob bytes

romamik:
FastMatrix.setFrom is useful when you want to write code like:

    transform = transform.multmat(FastMatrix3.rotation(node.rotation));
    ...
    g2.pushTransformation(transform);

With this code haxe fails to inline constructor, but with

    transform.setFrom(transform.multmat(FastMatrix3.rotation(node.rotation)));
    ...
    g2.pushTransformation(transform);

 - no allocation of heap objects occur.

As for Blob.bytes in most backends Blob.bytes is public and of type Bytes, so I made it more consistent.

### DIFF
--- a/Backends/Flash/kha/Blob.hx
+++ b/Backends/Flash/kha/Blob.hx
@@ -5,11 +5,16 @@ import haxe.io.Bytes;
 import haxe.io.BytesData;
 
 class Blob implements Resource {
-	public var bytes: BytesData;
+	var bytesData: BytesData;
+	
+	public var bytes(get, never):Bytes;
+	function get_bytes() {
+		return Bytes.ofData(bytesData);
+	}
 	
 	@:allow(kha.LoaderImpl)
 	private function new(bytes: BytesData) {
-		this.bytes = bytes;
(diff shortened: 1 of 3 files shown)


RobDangerous:
Hey, will gladly take the FastMatrix3 change (can you send another pull request for that?) but can not take the Blob change - bytes is not public because Kha explicitly does not guarantee that a Blob is based on a Bytes object (and for example in https://github.com/KTXSoftware/Kha/blob/master/Backends/Node/kha/Blob.hx it actually isn't).

sh-dave:
@romamik i don't know what your actual usecase is (f.ex. if you need random access to data etc), but i solved this issue in our framework by using a simple haxe.io.Input wrapper:

```haxe
class BlobInput extends haxe.io.Input {
	var blob : kha.Blob;
	var position = 0;

	public function new( blob : kha.Blob ) {
		this.blob = blob;
	}

	override public function readByte() : Int {
		try {
			return blob.readU8(position++);
		} catch (e : Dynamic) {
			throw new haxe.io.Eof();
		}
	}
}
```

PR #345, opened 2016-07-19
Added Graphics2.drawImageNoTransform() function

romamik:
Added Graphics2.drawImageNoTransform() function, that draws image with manually set vertex and texture coordinates and do not apply transformation. This can be used for example for sprite meshes.


romamik:
No need for this. We switched to g4 and this is the best choice if you need something like this.

PR #367, opened 2016-08-10
System.screenDpi() function. 

romamik:
Added System.screenDpi() function. Implemented for android, android-native, flash, html5 and windows. Returns 0 on other platforms. Relies on my pull request to Kore: https://github.com/KTXSoftware/Kore/pull/80.


romamik:
build fails, because it needs to update kore to my version, which is not present in your repository...

PR #376, opened 2016-08-13
Fix canvas size and enable/disable webgl

hammeron-art:


RobDangerous:
Sorry, I can't take the canvas fix. The canvas size isn't changed from inside Kha on purpose for easier embedding.

hammeron-art:
I see. It makes sense.

RobDangerous:
Thanks :)

RafaelOliveira:
@hammeron-art There are many things that happens in a page, so it's better to let the developer manage the canvas, it's more flexible.

hammeron-art:
@RafaelOliveira 
However it don't appears to have a easy way to change it at runtime.
Unless you do something like this:
```
@:privateAccess {
	var g:kha.js.CanvasGraphics = cast framebuffer.g2;
	g.width = 960;
	g.height = 540;
	g.canvas.width = 960;
	g.canvas.height = 540;
	g.canvas.save();
}
```

RafaelOliveira:
@hammeron-art 
You can see here a method that I use to set the canvas to use all the screen available in the browser (the code is similar):
https://github.com/RafaelOliveira/GameOfLife/blob/master/Sources/Main.hx#L54
This is before System.init

To made the change during the game you also need to call framebuffer.g4.viewport (if is webgl). 
I have a library that manage the canvas for html5 games that runs on mobile browsers using the entire screen. When the screen is rotated the size of the browser changes and the library change the canvas together. You can see here with a example:
https://github.com/RafaelOliveira/Mobile

PR #386, opened 2016-08-20
Android onCreate and onStart callbacks.

caiopsouza:
Created callbacks for android's main activity's onCreate and onStart.


RobDangerous:
Looks perfect, thanks!

PR #393, opened 2016-08-28
Graphics2 redesign

hammeron-art:
Style parameters, fill and stroke, renaming, batching improved for g4 implementation.

An example using most of the features:

``` haxe
var date = Date.now();
var hour = date.getHours();
var minutes = date.getMinutes();
var seconds = date.getSeconds();

var width:Float = kha.System.windowWidth();
var height:Float = kha.System.windowHeight();
var centerX = width / 2;
var centerY = height / 2;
var radius = Math.min(width, height) / 2;
var image = kha.Assets.images.sprite;

var g = framebuffer.g2;

g.style.font = kha.Assets.fonts.notoSansRegular;
g.style.fontSize = 32;
g.style.fillColor = kha.Color.Magenta;

var style1 = new Style();
style1.fill = false;
style1.strokeColor = kha.Color.Cyan;
style1.strokeWeight = 3;

var style2 = new Style();
style2.fillColor = kha.Color.White;
style2.strokeColor = kha.Color.Orange;
style2.strokeWeight = 1;

var style3 = new Style();
style3.strokeColor = kha.Color.Red;
style3.strokeWeight = 2;

g.begin(true);

// Frame
g.push();
g.translate(centerX, centerY);

// Images
g.image(image, -128, 16);
g.scaledImage(image, -256, 64, 64, 64);
g.subImage(image, -256-64, 128, 120, 100, 120, 100);
g.scaledSubImage(image, -256-128, -64, 120, 100, 120, 100, 128, 128);

g.ellipse(0, 0, radius, radius, style1);
var start = g.pop();

// Pointers
g.translate(centerX, centerY);
g.ellipse(0, 0, 32, 32, style1);

g.rotate(-Math.PI / 2);

style2.fillColor = kha.Color.fromFloats(1.0, 1.0, 1.0, 0.5);
g.push();
g.rotate((hour / 12) * Math.PI * 2);
g.triangle(32, -4, radius/2, 0, 32, 4, style2);
g.pop();

g.push();
g.rotate((minutes / 60) * Math.PI * 2);
g.rect(32, -4, radius-64, 8, style2);
g.pop();

g.push();
g.rotate((seconds / 60) * Math.PI * 2);
g.line(32, 0, radius-32, 0, style3);
g.pop();

for (i in 1...13) {
    g.rotate((Math.PI * 2) / 12);
    g.text('$i', radius - 16 - g.style.font.width(g.style.fontSize, '$i') / 2, -g.style.font.height(g.style.fontSize) / 2);
}

// Misc
g.setTransform(start);

var trans = kha.math.FastMatrix3.rotation((Math.PI * 2 / 12) * (seconds % 12));
var p1 = new kha.math.FastVector2(radius-48, 0);

g.beginShape(kha.graphics2.Primitive.Lines, style1);
for (i in 0...12) {
    p1 = trans.multvec(p1);
    g.vertex(p1.x, p1.y);
}
g.endShape(true);

g.end();
```


RobDangerous:
Oh, wow. Will go through it when I'm back home (mid next week).

dstrekelj:
Nice pull request! The styles in particular look very interesting!

As a user, though, I want to voice my concerns about some of the Graphics2 API changes.

Clarifying method arguments is always a good move, there's no harm in that. But a change to method identifiers is a _breaking change_.

In my opinion, omitting the 'draw' prefix from method identifiers is not reasonable. I do not see how there is anything to gain from this change other than to save four keystrokes. And even that is debatable because of code-completion.

The change also seems inconsistent. Not all methods were renamed (`drawVideo()`, for example), and `graphics4.Graphics2` remained untouched. This might unintentionally lead to more confusion.

All of the above leads me to believe that, in the case of this change, compatibility with existing Kha projects will be broken _for no apparent benefit_.

Again, this is only my opinion. It might be completely wrong, so I'd love hear your thoughts and arguments why the API changes were necessary. Maybe there's something I'm missing.

Sources/kha/graphics4/Graphics2.hx
```diff
@@ -270,17 +270,15 @@ class ColoredShaderPainter {
...
 	
-	private static var bufferSize: Int = 100;
-	private var bufferIndex: Int;
-	private var rectVertexBuffer: VertexBuffer;
-    private var rectVertices: Float32Array;
+	private static var maxVertexSize: Int = 1024;
+
+	private var vertexBuffer: VertexBuffer;
+	private var vertexIndex: Int;
+    private var vertices: Float32Array;
```
jamiltron: Is this spacing intentional?

Backends/HTML5/kha/js/CanvasGraphics.hx
```diff
@@ -92,7 +97,7 @@ class CanvasGraphics extends Graphics {
 		canvas.globalAlpha = 1;
 	}
 	
-	override public function set_color(color: Color): Color {
+	/*override public function set_color(color: Color): Color {
```
jamiltron: Might as well remove it if you're going to comment it out.

Sources/kha/graphics4/Graphics2.hx
hammeron-art: Yes. I tend do separate chunks of code by purpose.

hammeron-art:
@dstrekelj 
I forget to refactor drawVideo. I have to remove the methods with "fill" prefix too because we have styles now.
Note that when you call these methods, you are not drawing. In most of the cases you are building a buffer to be draw when end() is called. They are methods to add vertices so I also thought of adding a Shape class. This way you can store the result and reuse it.
For these reasons, I see a lot of benefit. The prefix is misleading and the change gives room for more for improvement.

tienery:
> The prefix is misleading

I don't think the `draw` or `fill` is misleading. It's the accepted convention among most game programmers to assume such calls exist, since SDL has been using it since day one and pretty much standardised such naming conventions. Removing it I think would cause confusion, because it's unusual to think it would be anything else, especially since Kha is supposed to be low-level.

And by extension of the point I just made, while the Styles approach is certainly nice, I personally think that's better suited as part of a GUI library, perhaps.

You could take advantage of Haxe's extension by doing this:

```haxe

class GraphicsExtension
{

      public static function drawRectEx(g2:kha.graphics2.Graphics, [your additional arguments here])
      {
           //do stuff
      }

}

```

And then later, in use, people may optionally use that api by doing this:

```haxe
using GraphicsExtension;
```

And it would appear inside of the `Graphics` class as you placed it normally. I don't think changing the entire class is necessary to make that happen, which would otherwise break other peoples' projects when they upgrade.

>  In most of the cases you are building a buffer to be draw when end() is called.

This is also unnecessary. If the developer wants to make their application double-buffered, that should be entirely their choice, instead of it being forced upon them.

This is just a few pointers, I'm not suggesting you should close and scrap this idea entirely. Although, double-buffering is automatically handled in the graphics card, so in this case it might be triple buffering on some targets. Sure, that can be handy for some systems, but again that should be a choice given to the developer.

wighawag:
In my opinion the use of style parameter is a better match for a low level library as it simplify reasoning about the graphics state (by removing it). 
We could even get rid of  the style property of Graphics. 

A library that made good of that style of api is skia :  https://skia.org/

Note that a library could be build on top to support the old style.

nanjizal:
The pull request does not provide vector hitTest or image/gradient fill because the base drawing object do not feature uv and color control of each corner of a triangle in the way that Trilateral/TrilateralXtra is currently evolving. GraphicsExtensions and scaledSubImage, are probably too limiting and I am now moving to provide methods over the triangle base to make more flexible versions of subImage drawing but it's not easy to get right. I would like to explore a dual Image shader with 3 corner separate uv and colours for aspects like drawing brushes, I am starting to look at caching upscaled vector groups.

My worry is that this pull could actually limit Kha and does not really provide the features needed, which may actually be way beyond any of our current perceptions - even though it looks very cool and I am very aware of how much work this might have involved because I am working on similar problems. I would welcome help with improving Trilateral or discussing ideas further, and I don't wish to offend in my critic of the solution presented, it's a very difficult solution space.

RobDangerous:
The relevant metric is not how many features can be put into g2, it's how well it maps to the supported and potential future g2 backends (<canvas>, Java Painter, WPF whatever) and how fast g4/g5 implementations can run it. Hit tests are definitely irrelevant (also no idea how color values could help with those), not sure about gradients but feel free to look that up in the mentioned APIs.

tienery:
The only way to efficiently draw gradients is by using shaders as the graphics card does colour transforming automatically. Because that's the only way to efficiently draw them, you would have to wrap g4 calls into the g2 API, which is what I think it's doing anyway but you would need to play around with it.

It does mean software rendered targets will be left out. The only way to do it is by using a bitmap with a colour mask, but that's obviously less efficient. It might be alright in the software rendering world, but I've never tried it.

RobDangerous:
Nah, gradients in a software renderer aren't much of a problem, many 90s 3D games did color interpolation calculations for every screen pixel even before 3D hardware. There are no software rendered targets though.

nanjizal:
tienery
The approach I have taken is to build a complex gradient with multiple triangles, it uses the standard shader used in Graphics2 the result can be used as a texture, so it's not a software renderer you don't need to play with g4 it works more or less like g2 implemention but with revised api.
https://nanjizal.github.io/TrilateralBazaar/linearGradient/build/html5/index.html
internals
https://github.com/nanjizal/Trilateral/blob/master/src/trilateral/tri/TriangleGradient.hx#L67

Robert
In terms of my approach working with other tech targets since it's based on Graphics2 internals then it will work with all targets accept the discussed limitations of Canvas which are approachable:
https://github.com/nanjizal/canvasImageTriangle
I have not looked at how g2 changes for g5 but would be curious.

PolyPainter has been adapted for NME it was not really planned but it was quite easy, the concepts prior to implementing in Kha was prototyped in webgl although in a nastier form, so I do not see the technical issues with future targets, although the design is not yet polished and I am still changing api aspects. NME version here:
https://github.com/nanjizal/TrilateralXtra/blob/master/src/trilateralXtra/fDrawing/PolyPainter.hx
simple use
https://github.com/nanjizal/nmeTriangleTest/blob/master/src/Sample.hx
more complex use
https://github.com/nanjizal/nmeTriangleTextTest/blob/master/src/Sample.hx
and in kha it's the same but you get the beauty of g.transformations :) and it's probably faster.
https://github.com/nanjizal/nmeTriangleTextTest/blob/master/src/Main.hx

Trilateral currently supports non image fills via helpers so pixels, canvas, swing
https://github.com/nanjizal/pixelDrawing
https://github.com/nanjizal/justDrawing
use 
https://github.com/nanjizal/hxbitTriangleTest/blob/master/src/Main.hx
and could easily support c#, in much the same way as Lime, there are details on Cambia github with music notation on use of c# and it looks fine. But the point being the triangle approach is flexible enough to support multitude of targets that Kha in future wants, but as I said I was not offering my solution more pointing out that the proposed one seemed too limited in flexibility, my library would ignore it and perhaps be hampered by it.

Jigsawx was originally developed with a Swing implementation which is like an early justDrawing mentioned above so while I am unsure on uv triangle support certainly arbitary image shapes can be created, having explored some lwjgl/slick I am unsure on which aspect of Java is going to be a problem?

**HIT TEST** let me explain.
hitTests in trilateral is easy since drawing is done with triangles any shape is an array of triangles the hitTest is fast, further the bounds are also saved as the triangles are created ( polypainter needs a new option for drawing to generate triangles where the source is not triangles and a hit is needed but often it is fed triangles ).
https://github.com/nanjizal/Trilateral/blob/master/src/trilateral/tri/Trilateral.hx#L71
https://github.com/nanjizal/Trilateral/blob/master/src/trilateral/tri/Trilateral.hx#L78
https://github.com/nanjizal/Trilateral/blob/master/src/trilateral/tri/TriangleArray.hx#L76
https://github.com/nanjizal/Trilateral/blob/master/src/trilateral/path/Base.hx#L23

If an image is renderTexture cached, and then scaled or moved - then when drawn as two new triangles with a g.transformation then the transformation can be applied to the mouse and passed to the hitTest, details vary but something like this:
```
var mouseNew = shape.transform.multvect( new FastVector2( mouseX, mouseY ) );
var hits = triangles.fullHit( mouseNew.x, mouseNew.y );
var selected = ( hits.length > 0 ) haxe.ds.ArraySort.sort( hits, sortDepth )[0]: null;
```
This allows fairly accurate vector shape mouse hits, as implemented in the jigsaw test. So generally it's often best to use transform with shape placement rather than x, y so that mouse can be mapped with scaling and rotation easily.

RobDangerous:
I wasn't talking about just making things run on other APIs - there need to be direct equivalents in those APIs or it's out of scope for g2 because g2 is just a portability layer for those APIs.
Your hit test answer doesn't explain why you would need color and/or uv coordinates and as I said, irrelevant for Kha.
Anyways, way out of topic by now. I like this pull request, precisely because it's not about adding features - just need to find the time to look at it in detail which clearly isn't easy for me.

RobDangerous:
Sorry, but keeping thins as they are. The usual recommendation nowadays is to copy graphics4.Graphics2 and adjust what you want when you want 2D things to be different.

PR #399, opened 2016-09-05
First work on enabling generic buffer support in compute shaders.

AoE-Maniac:
Uploading as a discussion base since I am unsure whether modifying vertex data in a compute shader requires major api changes:
Could be that the concept of another buffer (ShaderStorageBuffer) is flawed and we need a modified VertexBuffer instead. However, this seems incompatible with having multiple elements (e.g. pos and texCoords) in asingle VertexBuffer when only one is used in the compute shader.


RobDangerous:
Don't know about vertex buffer usage yet but apart from the index which I'll replace with a BufferLocation we can use this as it is.

PR #406, opened 2016-09-16
Support local XMLHttpRequests.

luboslenco:
I am loading blobs from local file system using the '--allow-file-access-from-files' chromium flag. However, these seem to return status of 0 and empty statusText on success.

This should not affect existing blob loading, and will only pass for local files when mentioned flag is enabled. Let me know what do you think.


RobDangerous:
I think that makes sense.

PR #410, opened 2016-09-24
Keep KhaActivity and Assets when using DCE

caiopsouza:


caiopsouza:
I should've explained a little better. When using DCE if you don't keep KhaActivity you get a black screen. If you don't keep the assets, you get a null pointer exception when trying to use them.

tienery:
I'm assuming this commit is supposed to fix the bug I'm currently experiencing (and yes, I am using the latest kha):

http://imgur.com/fzedG8S

I've checked my settings and they appear to be pointing in the right direction:

```js
// Place your settings in this file to overwrite the default settings
{
     "kha.khaPath": "C:/HaxeToolkit/haxe/lib/kha/git"
}
```

caiopsouza:
This only affect the android+java backend.

Join us on IRC if you need any help. #kha channel at ktxsoftware.com﻿

PR #415, opened 2016-09-30
Document realTime vs. time in Scheduler

ashes999:
This is a documentation change only.

### DIFF
--- a/Sources/kha/Scheduler.hx
+++ b/Sources/kha/Scheduler.hx
@@ -291,10 +291,18 @@ class Scheduler {
 		#end
 	}
 
+	/**
+	 * The amount of time (in fractional seconds) that elapsed while the game was active.
+	 * Doesn't include time when the game loses focus, until it regains focus.
+	 */
 	public static function time(): Float {
 		return current;
 	}
 	
+	/**
+	 * The amount of time (in fractional seconds) that elapsed since the game started.
+	 * This value keeps increasing even when the game loses focus.
+	*/
 	public static function realTime(): Float {
 		return System.time - startTime;
 	}


ashes999:
This doesn't seem to be the actual time difference. I don't know how I got these two values to be different. I'll close this until I can figure out what the actual difference is (reading the code doesn't make it obvious to me).

dstrekelj:
From what I understand, `realTime` is the time that has passed between an observed point in time and the point at which the Kha application started running. Basically, it's the running time of the application.

While `realTime` and `time` are similar, the key difference is that `time` does not have to constantly propagate forwards. This is because the scheduler is capable of "going back " to a certain point in time to re-execute scheduled time tasks. The change in time is reflected by the `time` variable, while the `realTime` one keeps chugging along as it always was.

Therefore, `time` can be thought of as the scheduler's own time, while `realTime` is the overall application running time. This means that `realTime` should always be either equal to or greater than `time`, since the latter can be moved back to a past time in the application lifetime.

RobDangerous:
Not quite... time is simulation time, all kinds of things can happen to it. In multiplayer games it can jump in all directions to synchronize clients, in single player games it is adjusted to provide a smooth experience and can therefore also move in both directions.

ashes999:
Can we reject this PR and can someone propose an actual, accurate, succinct explanation of time vs. realTime?

I experimented with an HTML5 app. When I alt-tab, wait, and alt-tab back, `time` freezes for that interval, while `realTime` doesn't.

tienery:
So, basically `time` is used for measuring in-game time while `realTime` is measured based on the system clock. I use `time` for calculating delta.

So perhaps this should be documented:

> Calculates the time, since the beginning of the game, based on frames per second.

> Calculates the time, since the beginning of the game, based on the system clock.

If I am understanding it correctly.

RobDangerous:
Merged and edited.

ashes999:
Thanks, much appreciated.

PR #422, opened 2016-10-11
A few Android externs

caiopsouza:


caiopsouza:
Some files had "\r" as line ending. That's why the diff is weird in some places.

PR #429, opened 2016-10-23
Fix crash on Chrome when local storage is not available

caiopsouza:
Chrome has an option to block third-party from accessing local storage and throws an exception when trying to. This prevents the script from crashing.


wighawag:
Would be great to either have a function to check if local storage is available, or alternatively/additionaly we should make "write" return a boolean so we can track success or failure.

caiopsouza:
I agree. Silently failing is not a good idea.

My biggest concert was the game crashing, but eventually I'll have to implement something to inform the player that the game cannot be saved.

Issue #219, opened 2016-01-23
Load a image from the web


RafaelOliveira:
I'm trying to load a png file from the web and create a Image with the data. I made some code with the help of Robert, but I don't know much about read bytes from a image. I'm putting my code here so other people can help and the result can be merged in Kha.
The Format library is needed for this.

This is the last result with a black ball. It's almost there, but it's not the image.

![image](https://cloud.githubusercontent.com/assets/69374/12531431/70e667e8-c1e1-11e5-9d15-ea197c7b8b1a.png)

the code:

``` haxe
package;

import haxe.io.Bytes;
import haxe.io.BytesInput;
import kha.Color;
import kha.Framebuffer;
import kha.Image;
import kha.Scaler;
import kha.System;
import format.png.Reader;
import format.png.Data;
import format.png.Tools;

class BasicKha 
{
    var backbuffer:Image;   
    var urlImage:Image; 

    public function new() 
    {
        backbuffer = Image.createRenderTarget(800, 600);        

        // note that in html5 we need to deal with the 'Access-Control-Allow-Origin' problem
        // so this function is more usable in cpp targets

        loadImageFromUrl('http://sudoestegames.com/play/ball.png', function(image:Image) {
            urlImage = image;           
        });
    }   

    public function update():Void
    {


    }

    public function render(framebuffer:Framebuffer):Void 
    {       
        backbuffer.g2.begin(true, kha.Color.White); 

        if (urlImage != null)
            backbuffer.g2.drawImage(urlImage, 10, 10);      

        backbuffer.g2.end();

        framebuffer.g2.begin();
        Scaler.scale(backbuffer, framebuffer, System.screenRotation);
        framebuffer.g2.end();
    }

    function loadImageFromUrl(url:String, callback:Image->Void):Void
    {       
        var http = new haxe.Http(url);
        http.onData = function(strData:String)
        {           
            var bytes = Bytes.ofString(strData);            
            var reader = new Reader(new BytesInput(bytes));
            var data = reader.read();
            var header = Tools.getHeader(data);

            var tempImg = Image.create(header.width, header.height);
            var pngbytes = Tools.extract32(data);
            var imagebytes = tempImg.lock();

            for (y in 0...tempImg.height) for (x in 0...tempImg.width) 
            {
                imagebytes.set(y * Std.int(tempImg.width) * 4 + x * 4 + 0, pngbytes.get(y * Std.int(tempImg.width) * 4 + x * 4 + 2));
[...]

PR #440, opened 2016-11-07
Image.fromFileBytes

romamik:
Function to create Image from file loaded from some external source. Implementation for Html5, Flash and Kore.

### DIFF
--- a/Backends/Flash/kha/Image.hx
+++ b/Backends/Flash/kha/Image.hx
@@ -44,6 +44,16 @@ class Image implements Canvas implements Resource {
 	public static function fromBytes(bytes: Bytes, width: Int, height: Int, format: TextureFormat = null, usage: Usage = null): Image {
 		return null;
 	}
+	
+	public static function fromFileBytes(bytes: Bytes, fileExtention: String, doneCallback: Image -> Void, errorCallback: String->Void): Void {
+		function handleError(e:flash.events.Event) errorCallback(e.toString());
+		var loader = new flash.display.Loader();
+		loader.contentLoaderInfo.addEventListener(flash.events.Event.COMPLETE, function(_) doneCallback(Image.fromBitmap(loader.content, false)));
+		loader.contentLoaderInfo.addEventListener(flash.events.IOErrorEvent.IO_ERROR, handleError);
+		loader.contentLoaderInfo.addEventListener(flash.events.SecurityErrorEvent.SECURITY_ERROR, handleError);
+		loader.contentLoaderInfo.addEventListener(flash.events.AsyncErrorEvent.ASYNC_ERROR, handleError);
+		loader.loadBytes(bytes.getData());
+	}
 
 	public function new(width: Int, height: Int, format: TextureFormat, renderTarget: Bool, depthStencil: DepthStencilFormat, readable: Bool) {
 		myWidth = width;
--- a/Backends/Flash/kha/Image.hx
+++ b/Backends/Flash/kha/Image.hx
@@ -45,7 +45,7 @@ class Image implements Canvas implements Resource {
 		return null;
 	}
 	
-	public static function fromFileBytes(bytes: Bytes, fileExtention: String, doneCallback: Image -> Void, errorCallback: String->Void, readable:Bool = false): Void {
+	public static function fromEncodedBytes(bytes: Bytes, fileExtention: String, doneCallback: Image -> Void, errorCallback: String->Void, readable:Bool = false): Void {
 		function handleError(e:flash.events.Event) errorCallback(e.toString());
 		var loader = new flash.display.Loader();
 		loader.contentLoaderInfo.addEventListener(flash.events.Event.COMPLETE, function(_) doneCallback(Image.fromBitmap(loader.content, readable)));
--- a/Backends/Flash/kha/Image.hx
+++ b/Backends/Flash/kha/Image.hx
@@ -45,10 +45,10 @@ class Image implements Canvas implements Resource {
 		return null;
 	}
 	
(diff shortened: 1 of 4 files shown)


romamik:
Checks fail because need Kore version that exists only in my fork.

sh-dave:
Why is it called fromFileBytes; shouldn't matter if it's loaded from some file or a buffer created in memory imo? The ```fileExtension``` parameter seems strange as well. You basically just pass it around. Can't you just bind that to the error callback you're registering?

romamik:
>Why is it called fromFileBytes; shouldn't matter if it's loaded from some file or a buffer created in memory imo? 
Yes, that's not perfect name. I just failed to think of better. There is already function with name fromBytes() which accepts raw image data, already decoded from image file format. So to distinguish from this function I used this name. 
If somebody have better name for this function - I'm ready and will be glad to rename.

>The fileExtension parameter seems strange as well. You basically just pass it around. 
On flash and Html5 it is not used, but it's used in Kore: it uses file extension to determine the file format.
I think it's possible to make Kore determine file format based on contents of the file rather then the file extension, but this will require hard testing. You can never know where it will break after such changes...

>Can't you just bind that to the error callback you're registering?
Sorry, but I did not understand what you mean.

sh-dave:
If ```fromBytes()``` is already decoded, then ```fromUndecodedBytes()``` or ```fromRawBytes()``` maybe?

And forget what i said about the parameter, i missed the format check in Kore and thought it's only used for the error callback.

romamik:
In my opinion raw bytes are accepted by fromBytes, not by my function. Also it's not correct to say that fromBytes accepts uncompressed bytes, because they not nececary were ever compressed. 

May be we can rename fromBytes to fromRawBytes, but this is a breaking change, so no, this is not good. May be we can call my fuction fromCompressedBytes, even if they are not necesary compressed.

romamik:
I rebased this pull request...

romamik:
I've decided to have this code external to kha: https://gist.github.com/romamik/cefe4ad27ffd0da3b7b02985b12233fb
Closing this pull request.

RobDangerous:
I might still resurrect this if it's ok for you, still have to have a closer look.

romamik:
Yes, I can reopen this if you think so.
The main reason why I reimplemented this as external code is that it's easier for me to maintain my project up to date with latest kha in this way.

Also I feel that Kore implementation is a bit hacky. Right way is to make Kore::Image constructor accept Reader instead of filename, but this involves changing Texture because it is inherited from Image, and which is platform dependent, so the changes will be in every target. And I do not have possibility to test every target.

RafaelOliveira:
This is very useful, because can be used to resolve https://github.com/Kode/Kha/issues/219 (load a image from web), together with haxe.Http. I tested it now in Windows, and it worked. For me this is very important.

sh-dave:
I also think this would be a very welcome addition.

RobDangerous:
Added a non-hacky Kore implementation. Please test because I didn't.

romamik:
>Please test because I didn't.
:))) you are real jedi!!!

31 янв 2017 г. 5:55 PM пользователь "Robert Konrad" <
notifications@github.com> написал:

Added a non-hacky Kore implementation. Please test because I didn't.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/Kode/Kha/pull/440#issuecomment-276384692>, or mute the
thread

RafaelOliveira:
It works for Windows, but not for html5. I'm testing with haxe.Http, and converting the data with Bytes.ofString() (Http receives the bytes as string). The bytes are being received, but Image.fromEncodedBytes doesnt work and call the errorCallback.

To test with Http this need to be tested in KodeStudio because Electron doesnt have crossorigin restrictions.

RobDangerous:
No idea whether haxe.Http is the right thing to use for binary files.

romamik:
Yes, haxe Http class is not suited for binary files. It by chance works for sys (c++ and neko) target, but for js and flash it just do not work. I have the class that handles this, I can't make it open source, but basic idea is very simple:
```haxe
    class BinaryHttp {
        getBytes(url:String):Promise<Bytes> {
           var deferred:Deferred<Bytes> = new Deferred();
           #if js 
		request = Browser.createXMLHttpRequest();
		request.onload = fucnction() deferred.resolve(Bytes.ofData(cast(request.response, BytesData)));
		function onError() throw 'Error';
		request.onerror = onError;
		request.onabort = onError;
		request.responseType = ARRAYBUFFER;
		request.open('get', url, true);
		request.send();
           #elseif flash
                //var loader = new URLLoader();
                //loader.dataFormat = URLLoaderDataFormat.BINARY;
                #error
           #elseif sys
                var http = new Http(httpBytes.url);
                http.onError = function(_) throw 'error';
                http.onData = function(data:String) deferred.resolve(Bytes.ofString(data));
                http.request();
           #else
               #error
           #end
           return deferred.promise();
        }
    }
```
This code was not tested, but gives the idea.

romamik:
Also, I must say, that if you really need to download file from web and show, then downloading it like that and showing using code from this pullrequest makes some overhead: image is downloaded as byte array, encoded to base64, then decoded from base64, then decoded from image file format, and only then uploaded to videomemory. While it's completely possible to avoid encoding/decoding from base64: you need to create  element with src=url of picture and then create kha image from element (look at the code from pullrequest, it does exactly this, just with data url).

RafaelOliveira:
For html5 Assets.loadImageFromPath() already works, and it creates a image element, so the use of Http in html5 was just for testing.

RobDangerous:
loadBlobFromPath might work for you then.

PR #457, opened 2016-12-08
return windowId again

sh-dave:
Some mighty evil :smiling_imp:  thought returning -1 as windowId is fun.


RobDangerous:
It isn't?

PR #466, opened 2016-12-22
html5: fix for kha.Image.fromBytes()

sh-dave:
Using kha.Image.fromBytes() explodes without the fix.

Now i'm not sure if kha.Image.fromImage() will break, as i have no idea on how to get an ImageElement to test. I think WebGLImage needs a bit of cleanup, especially the ```image: Dynamic``` so the correct texImage2d function can be called in createTexture() depending on what ```image``` actually is?

(http://api.haxe.org/js/html/webgl/RenderingContext.html#texImage2D)

(and sorry about the whitespace noise, i have autotrim in my IDE enabled)

### DIFF
--- a/Backends/HTML5/kha/WebGLImage.hx
+++ b/Backends/HTML5/kha/WebGLImage.hx
@@ -157,7 +157,7 @@ class WebGLImage extends Image {
 			default:
 				SystemImpl.gl.texImage2D(GL.TEXTURE_2D, 0, GL.RGBA, realWidth, realHeight, 0, GL.RGBA, GL.UNSIGNED_BYTE, null);
 			}
-			
+
 			if (format == DEPTH16) {
 				SystemImpl.gl.framebufferTexture2D(GL.FRAMEBUFFER, GL.DEPTH_ATTACHMENT, GL.TEXTURE_2D, texture, 0);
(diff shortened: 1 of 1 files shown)


RobDangerous:
ImageElements and Image.fromImage are used for all regular images, so I strongly suspect you accidentally tested that. As far as I can tell image is always an ImageElement, we should just type it properly.

RobDangerous:
Well, that didn't work, have to have a closer look after all.

sh-dave:
Well, it works for Image.fromBytes(), as this is actually an ArrayBufferView and not an ImageElement. Not sure what function @RafaelOliveira called that calls createTexture() in turn.

RafaelOliveira:
It was a simple code:

```haxe
class Project {
	public function new() {
		Assets.loadEverything(assetsLoaded);
	}

	function assetsLoaded()
	{
		System.notifyOnRender(render);
		Scheduler.addTimeTask(update, 0, 1 / 60);
	}

	function update(): Void {
		
	}

	function render(framebuffer: Framebuffer): Void {
		framebuffer.g2.begin(true, kha.Color.Black);
		framebuffer.g2.drawImage(Assets.images.citron, 20, 20);
		framebuffer.g2.end();	
	}
}
```

PR #484, opened 2017-01-10
Kore backend: respect texture->stride() and convert RGBA->BGRA if direct3d in Image.unlock()

romamik:
Not sure if you have to merge this pull request as it is: it fixes Image.lock()/unlock() for Kha, but IMHO it should be fixed in Kore. https://github.com/Kode/Kore/issues/134
I see different options:
1. Fix both in Kha - done here
2. Fix stride in Kha and color format in Kore
3. Fix both in Kore

### DIFF
--- a/Backends/Kore/kha/Image.hx
+++ b/Backends/Kore/kha/Image.hx
@@ -234,9 +234,13 @@ class Image implements Canvas implements Resource {
 
 	private var bytes: Bytes = null;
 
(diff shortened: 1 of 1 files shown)


RobDangerous:
I'll go for option 2 but will merge this for now (will remove the #ifdef DIRECT3D part when it's fixed in Kore).

PR #527, opened 2017-02-24
Image clear for Kore

luboslenco:
Adds a method to clear (non-rendertarget) textures to specified color. Should be especially useful for compute/imagestore/.. apis, once that gets settled in Kha.

To go with:
https://github.com/Kode/Kore/pull/150
https://github.com/Kode/Krom/pull/21

### DIFF
--- a/Backends/Android/kha/Image.hx
+++ b/Backends/Android/kha/Image.hx
@@ -387,6 +387,10 @@ class Image implements Canvas implements Resource {
 		
 	}
 
+	public function clear(x: Int, y: Int, z: Int, width: Int, height: Int, depth: Int, color: Color): Void {
+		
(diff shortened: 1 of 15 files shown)


OvermindDL1:
Wouldn't `fill` be a more usual name for this function?  Or at least as an alias of it?

RobDangerous:
Clear is clearly the name which is commonly used in modern gfx apis so I'll take it as it is.

luboslenco:
I would not mind changing the name if desired - I used clear to relate to [glClearTexImage](
 https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glClearTexImage.xhtml) which is handling this on OpenGL.

PR #573, opened 2017-04-30
Implement Shader.fromSource for Krom

luboslenco:
To go with https://github.com/Kode/Krom/pull/33.


RobDangerous:
Careful with Shader.fromSource, it's not portable and can't work at all on some targets.

Issue #556, opened 2017-04-10
[Request] Create screenshots and save as file


RafaelOliveira:
I'd like to save screenshots of the framebuffer in my HD. This would be useful to create videos from the files using some video editing app.

This gist creates screenshots in html5 but html5 is limited and can only shows in a new window.
https://gist.github.com/RafaelOliveira/050de54121b38b1945fb71624f258d3d

RobDangerous:
Lubos just added a getPixels function for render-targets.

RafaelOliveira:
I made a library, but the red colors are being saved as blue (both bmp and png). @RobDangerous do you know if this code is correct? I also tried png.tools.build32ARGB, but it generates images with the background transparent (but the background was painted black).

I adapted from this code http://stackoverflow.com/questions/14034644/save-an-array-of-pixels-to-a-file

```haxe
package;

import kha.Image;

class SaveFrames
{
	static var idx:Int;
	static var filePath:String;
	static var image:Image;

	public static function start(image:Image, startIndex:Int, filePath:String):Void
	{
		SaveFrames.image = image;
		idx = startIndex;
		SaveFrames.filePath = filePath;
	}

	public static function saveBmp():Int
	{
		#if cpp
		var handle = sys.io.File.write(filePath + idx + '.bmp', true);
		new format.bmp.Writer(handle)
		.write(format.bmp.Tools.buildFromBGRA(image.width, image.height, image.getPixels()));
		handle.close();
		#end

		idx++;
		return idx - 1;
	}

	public static function savePng():Int
	{
		#if cpp
		var handle = sys.io.File.write(filePath + idx + '.png', true);
    		new format.png.Writer(handle)
        	.write(format.png.Tools.build32BGRA(image.width, image.height, image.getPixels()));
    		handle.close();
		#end

		idx++;
		return idx - 1;
	}
}
```

RafaelOliveira:
I made a library that works (html5 and linux) based on https://github.com/Kode/Kha/pull/440#issuecomment-276461361. It will be put on the forum when finished.

RafaelOliveira:
I mixed things and closed the wrong issue.

RafaelOliveira:
The function getPixels is available to save a render target, so I closing this.

PR #575, opened 2017-05-04
Image.getPixels proposal

luboslenco:
I have experimented with several namings but so far this is the cleanest I think.
- lock()/unlock() to manipulate GPU data using CPU
- getPixels() to pull data from GPU, no upload

Alternative is to use .lock(read = true), but lock() implies I am also responsible to call unlock(). We would have to make sure unlock() does nothing in case of read = true.

I implemented getPixels() for RGBA32 render targets on WebGL, Krom and Kore.

To go with https://github.com/Kode/Kore/pull/175 and https://github.com/Kode/Krom/pull/35.

---

Also included for Krom (should have been separate):
- requestShutdown()
- fileSaveBytes() - mirrors haxe.io.File.saveBytes()
- sysCommand() - mirrors Sys.command()

### DIFF
--- a/Backends/Android/kha/Image.hx
+++ b/Backends/Android/kha/Image.hx
@@ -381,6 +381,10 @@ class Image implements Canvas implements Resource {
 		bytes = null;
 	}
 
+	public function getPixels(): Bytes {
+		return null;
+	}
+
 	public function generateMipmaps(levels: Int): Void {
 		
 	}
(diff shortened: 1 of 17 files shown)


RobDangerous:
Okidoki

RafaelOliveira:
From my tests https://github.com/Kode/Kha/issues/556 the red colors are being saved as blue (when saving to a file), but I don't know if is my code.

PR #576, opened 2017-05-05
Add Krom.savePath

luboslenco:
I think this should do for some spartan environment for making programs in Krom.

- Krom.loadBlob() to load files, which also handles absolute paths with https://github.com/Kode/Kore/pull/176
- Krom.fileSaveBytes() to write files
- Krom.sysCommand() to interact with OS
- Krom.savePath() to get a safe path for storing & parsing results of sysCommand() or app data

To go with https://github.com/Kode/Krom/pull/36.


RobDangerous:
We will foster an army of warrior programmers in our Spartan environment.

PR #580, opened 2017-05-10
More math functions

malublu:
What you think about it?
Not finished yet only a idea!

### DIFF
--- a/Sources/kha/math/FastMatrix3.hx
+++ b/Sources/kha/math/FastMatrix3.hx
@@ -169,4 +169,84 @@ class FastMatrix3 {
 			c20 * invdet,  -c21 * invdet,  c22 * invdet
 		);
 	}
+	
+	public inline function clone():FastMatrix3 {
+		return new FastMatrix3(_00, _10, _20, _01, _11, _21, _02, _12, _22);
+	}
+	
+	public static inline function identity():FastMatrix3 {
+		return new FastMatrix3(1, 0, 0, 0, 1, 0, 0, 0, 1);
+	}
+
+	public inline function set(x:Int, y:Int, value:FastFloat) {
+		switch(x) {
+			case 0:
+				switch(y) {
+					case 0: _00 = value;
+					case 1: _01 = value;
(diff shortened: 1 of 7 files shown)


sh-dave:
```haxe
public inline function translate(tx:FastFloat = 0, ty:FastFloat = 0):FastMatrix3 {
...
}
```
Don't use default  arguments, thats slow code on native and doesn't make sense to have in the first place.

---

Matrices already have ```identity()``` iirc.

---

```haxe
class FastVector2 {
...
    public inline function clone():FastFloat {
        return new FastFloat(x, y);
    }
...
}

class FastVector3 {
...
    public static inline function clone():FastVector3 {
        return new FastVector3(x, y, z);
    }
    
    public static function toFast():kha.math.FastVector3 {
        return new kha.math.FastVector3(x, y, z);
    }
...
}
```
Did you even try to use some of these functions? Thats broken just by reading it...

---

```haxe
public inline function set(x:Int, y:Int, value:Float) {
   ...nested switch nonsense...
}
```
Just access _00, _10 ... whatever directly

---

I'm sure there is a lot more i missed...

malublu:
Year I know.
That's because I said it's not finished yet. Only an idea.
But what's about the main idea with skew,rotate,translate functions?
The functions are only ported from my very old library (other language).

The set function are not really needed.
Sorry for that, but I only copy paste my code and modified the syntax, etc....
I would make a complete rework of the function.

Should the skew, rotate, etc.. function modify the matrix or should they create a new matrix?

It can be faster if you modify the matrix.
Like:

```
matrix.skew(...);
matrix.rotate(...);
```

Instance of:

```
// A matrix created, that you don't need (after skew)
matrix.skew(...).rotate(...);
```

sh-dave:
Current matrices are immutable, so i would keep it that way. They're usually inline so they don't actually allocate a new object at runtime. To modify the ```this``` matrix you can just call ```setFrom()``` with the result of the computation.

RobDangerous:
Sorry but let's keep it minimal in Kha itself.

PR #603, opened 2017-06-05
send directly the keyCode from the event in the key listeners

juakob:
send directly the keyCode from the event in the key listeners


RobDangerous:
Now I see what you meant. Does this actually work? Kha's key codes are based on DOM keycodes, no idea if they are related to Flash keycodes.

sh-dave:
yes, that should actually work. was about to do it like that myself, but then found this PR.

RobDangerous:
Cool

PR #605, opened 2017-06-06
GL_RED for single-channel float textures in WebGL2

luboslenco:
Otherwise Chrome is unhappy.


RobDangerous:
Oh no, what happened?

luboslenco:
I don't know, but the code code seems to be in:P
https://github.com/Kode/Kha/blob/master/Backends/HTML5/kha/WebGLImage.hx#L42

RobDangerous:
Pull-request-overflow

PR #617, opened 2017-07-02
Add tint Color.None = -1; to remove tint

Sondro:
Took me a few min to figure this out by doing a trace on a non-tinted sprite so I could remove the tint.


sh-dave:
Why no just use kha.Color.White?

Sondro:
Because Color.White should really be split into Color.WarmWhite and Color.CoolWhite

It's pretty much useless to tint 'White' right now.

RobDangerous:
Color.White is equivalent to your definition of None (because -1 is 0xffffffff in hex) and it should be used to avoid tinting. It is also the default color. Sorry for the trouble.

PR #623, opened 2017-07-17
Ints for Debug/Rebind

Sondro:
Give us back Ints for Debug/rebind without breaking anything. A lot of the basic KeyCodes are missing/won't work on Electron Debug on Win7.


sh-dave:
Why not fix the wrong / missing ones instead of adding something like that?
To check if you're running in electron you can use:

```haxe
#if kha_debug_html5
#end
```

RobDangerous:
I wouldn't even know how to use this, can you show some sample code for how to fix things with your ints?

Sondro:
When you trace a KeyCode it gives you the int, ie...

Tracing Equals for me gives '189', Equals is '61' in the current KeyCode.

'189' right now does not exist on the KeyCode list so you can't use it. 

I have another version with just the missing KeyCodes but then you have to look up the KeyCode list every time to know what KeyCode the Int is equal to instead of just putting 'i'+tracedInt for your KeyCode. If you just want to manage the Ints it makes more sense.

So for example then when rebinding/or alt keys you can do stuff like:

if(Keyboard.isPressed(KeyCode.i189) || Keyboard.isPressed(KeyCode.Equals)) //Equals varies per target

or...

if(Keyboard.isPressed(KeyCode.i189) || Keyboard.isPressed(KeyCode.i61) //Equals varies per target

This makes more sense when you want to have all the Keys Coded properly per platform without Kha having to know what each target's specifics are.

If we just had Ints with KeyCodes like before it would be even better because we wouldn't have to do any typecasting/conversion and could just enter the missing Ints that were traced and store/bind them however we want. 

Right now I'm using workarounds for the changes.

Before the changes I had stuff like...


if(Keyboard.isPressed(i189) || Keyboard.isPressed(KeyCode.Equals)) //Equals varies per target

or...

if(Keyboard.isPressed(189) || Keyboard.isPressed(61) //Equals varies per target

or...

var keyWinEquals = 189; 

if(Keyboard.isPressed(keyWinEquals)  //Equals varies per target

And also you can manage your own KeyCodes in Array<Int> which swap per platform or EVEN your own users bindings so WinKeyCode[x] or MacKeyCode[x] or PlayerRobsKeys[x], 'x' will always equal 'Equals' or your proper key because the Array just stores the Ints (traced or otherwise) in different orders/values. 

On top of this  OpenCurlyBracket and other keys that require shift don't work only have the base keys Ints in traces so you need to make your custom key combos with &&, although the previous system with characters like '}' for OpenCurlyBracket did work .

It all really depends on the architecture you want, but the current system of not being able to use basic keys like '=' and '-' which then also prevents  '_' and '+' is kinda a problem.

RobDangerous:
Well, the basic idea is that key codes should not vary by target - which is of course a work in progress but I do not want to incorporate temporary hacks like this - and if you really want to use them, I think you can just cast an Int to KeyCode. So, as Dave said, how about fixing the key codes instead?

sh-dave:
I wouldn't mind an implicit Int cast though for the time being.

```haxe
abstract KeyCode(Int) to Int {
...
}

function keydown( keycode: KeyCode ) {
   var asInt: Int = keycode;
}
```

And there do seem to be some diffrences/missing ones for me as well. These are codes from http://keycode.info (using chrome 59) for me (german microsoft natural QWERTZ keyboard):

```haxe
// DE / UK layout
	var DecreaseVolume = 174;
	var IncreaseVolume = 175;
	var Next = 176;
	var Previous = 177;
	var Stop = 178;
	var PlayPause = 179;
	var Email = 180;

	var Semicolon = 186; 
	var EqualSign = 187;
	var Dash = 189;

// UK layout only
	var BackTick = 223; // ` key left of 1; (it's code 192 on DE like current keycode)

// DE layout only
	var AngleBracket = 226; // key right of the left shift button; (it's code 220 on UK like current keycode)
```

Sondro:
Okay, I updated the code to just be like a base fix/template for generating if/ends with only the missing Ints. Does that work for now?

RobDangerous:
Sorry, I still don't get it. Why can't you just cast?

RobDangerous:
Now you don't even need to cast. One of the cool things about open source is by the way that you can always use your own fork so please feel free to do that while we fix it properly.

Sondro:
Thanks, appreciate the update.

I opened an issue and then later posted this hack to get attention.

Issue #453, opened 2016-11-22
dx, dy discrepancy 


jefvel:
There's an issue with the dx and dy values on mouse move in Chrome and Edge, http://i.imgur.com/s4xLNWY.gifv

However, it works properly in Electron and Firefox, 
http://i.imgur.com/vTk6MN6.gifv

The issue seems to be with when the screen size is not set to 100% in the display settings in Windows. When it's set to 100% it works the same way in chrome as in electron.

RobDangerous:
Hey, I think this is fixed.

PR #644, opened 2017-08-17
Fix mouse events in html

MarianThull:
If they are set, mouse position and delta is relative to width and height in SystemOptions. Resolves issue #453.

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -315,7 +315,7 @@ class SystemImpl {
 		//var heightTransform: Float = canvas.height / Loader.the.height;
 		//var transform: Float = Math.min(widthTransform, heightTransform);
 		if (gl) {
-			var g4 = gl ? new kha.js.graphics4.Graphics() : null;
+			var g4 = new kha.js.graphics4.Graphics();
 			frame = new Framebuffer(0, null, null, g4);
 			frame.init(new kha.graphics2.Graphics1(frame), new kha.js.graphics4.Graphics2(frame), g4); // new kha.graphics1.Graphics4(frame));
 		}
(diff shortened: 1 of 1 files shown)


MarianThull:
Thanks for your fast replies!

### About the first one:

The problem I was looking into with that part of the changes was that different browsers (for example Chrome vs. Firefox) showed different behaviour when the website is zoomed in or out.
Let's say our canvas is 100x100 px and you move your mouse halfway through it while on 50% zoom. Firefox will give you a _movementX_ of 50, while Chrome returns 25. But you are definitely right, it won't work for a locked mouse at the moment. I can look into that on Monday.

### About your second reply:

If you define _width_ and _height_ when calling _System.init_, use a render target in the same size and the canvas is resized for some reason, the rendered area will scale but the aspect ratio will not adapt. Instead black bars are added left and right or top and bottom. The changes I made would make sure that _x_ and _y_ of the mouse events are correct with regard to the rendered area.
rect
The rectangle in the picture is drawn with _drawRect(0, 0, width, height)_. With the suggested changes the mouse _x_ and _y_ at the top left would be (0, 0), at the moment it's (size_of_black_bar, 0). Additionally the _y_ coordinates at the bottom line would be _y=height_ instead of _height * some_scale_.

By "that were relying on these options not being taken into account", do you mean not being taken into account even when they are set? Because the behaviour when they are not set stayed the same.

wighawag:
>The problem I was looking into with that part of the changes was that different browsers (for example Chrome vs. Firefox) showed different behaviour when the website is zoomed in or out.
Let's say our canvas is 100x100 px and you move your mouse halfway through it while on 50% zoom. Firefox will give you a movementX of 50, while Chrome returns 25. But you are definitely right, it won't work for a locked mouse at the moment. I can look into that on Monday.

I personally use movementX only for lockMouse mode and I compute the diff myself for all the other case. What we could do is set movementX and movementY to zero when not in lockMouse mode so that these cannot be misused (assuming we cannot solve the cross browser issue) 

> The rectangle in the picture is drawn with drawRect(0, 0, width, height). With the suggested changes the mouse x and y at the top left would be (0, 0), at the moment it's (size_of_black_bar, 0). Additionally the y coordinates at the bottom line would be y=height instead of height * some_scale.

How did you get to that? is the black area, the canvas ?
Did you have any transform ?
A test app would be good to see

I have been using the mouse coordinate on a web page that support any size/ratio without any problem.


> By "that were relying on these options not being taken into account", do you mean not being taken into account even when they are set? Because the behaviour when they are not set stayed the same.

I meant "not being taken into account even when they are set". Icurrently use these options for native target but since on html5, there is no way nor desire to control the width and height of the user.

MarianThull:
> Did you have any transform ?

Thanks, I found my mistake. I am using the Scaler to draw my backbuffer onto the canvas. It seems like all I tried to "correct" were the transformations made by it.
You can see my code [here](https://github.com/MarianThull/KhaDev/tree/html-mouse-delta).

I will look into the cross browser issue with movementX some more, but you can forget about the other changes.

wighawag:
Cool.

I do not know how slow is it but I suppose the isFirefox check should be done once and stored in a variable so that we do not ask the browser every frame

MarianThull:
Oh, of course. Thank you!

Issue #146, opened 2015-08-28
drawString alignment option


wighawag:
Currently it seems drawString can only draw text top/left aligned to the position specified. It would be very useful if we could specify alignment both vertically and horizontally

RblSb:
Fixed

PR #663, opened 2017-10-02
Added string and character alignment options

FacelessJ:
Added horizontal and vertical alignment options to drawString and drawCharacters. As strings are single line, alignment is anchored to the top left corner of the drawn string (which includes a gap proportional to font size). This means drawing three strings at the same location with TextTop, TextMiddle and TextBottom vertical alignment (and same font size) will not overlap each other and have equal spacing between the three lines.

Addresses issue #146

### DIFF
--- a/Backends/HTML5/kha/js/CanvasGraphics.hx
+++ b/Backends/HTML5/kha/js/CanvasGraphics.hx
@@ -5,8 +5,6 @@ import kha.Color;
 import kha.FontStyle;
 import kha.graphics2.Graphics;
 import kha.graphics2.ImageScaleQuality;
-import kha.graphics2.HorTextAlignment;
-import kha.graphics2.VerTextAlignment;
 import kha.Kravur;
 import kha.math.FastMatrix3;
 import kha.math.Matrix3;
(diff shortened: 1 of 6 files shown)


wighawag:
optional arguments should be avoided performance wise, it creates extra allocation on hxcpp if I remember correctly. You could create an extra function drawAlignedString...

Also maybe this need not to be in Kha. I have coded the same functionality as part of a utility function.

FacelessJ:
Ah ok, I thought that was just for default arguments. I've fixed it up by introducing drawAlignedString and drawAlignedCharacters as suggested

RobDangerous:
I like it but I have to meditate on whether this should be in Kha or not.

tienery:
@wighawag it's interesting that the OP posted a link to an issue made by you two years ago asking for this feature, and now you're turning around and disagreeing with it? May I suggest closing the original issue, if that's the case.

Additionally, may I also suggest that text alignment should be part of a higher-level library than as part of the core framework. Kha sets itself as low-level, so I have the same thoughts as @RobDangerous 

You can also use the [Extensions](http://haxe.org/manual/lf-static-extension.html) class capabilities of Haxe.

FacelessJ:
Thanks for the advice. 

Just to clarify, would implementing drawAlignedString in StringExtensions.hx (or similar) count as moving it out of "core Kha", or do you mean a completely separate library built on top of Kha? Thanks.

tienery:
Well, StringExtensions.hx exists because `drawCharacters`, no other reason. Don't see it being useful in any other context, unless there's another library out there using this kind of feature...

And I do mean a separate library, which should be nothing to do with Kha. I'm sure that's what Rob is suggesting. I would, personally. Text Alignment would best fit inside a GUI library built on-top of Kha.

FacelessJ:
That makes sense.

Well, I'll leave it here for Rob to make a decision.

sh-dave:
I do see some merit in having it in Kha, and, we also have this class: https://github.com/Kode/Kha/blob/master/Sources/kha/graphics2/GraphicsExtension.hx

If possible (i haven't looked at the code), i would leave G2 untouched and implement it here.

wighawag:
I guess I changed my mind a bit. it was still a maybe though :)

I ll leave the issue open as my previous self and others seem to think that it might be good to have in Kha.

One thing I realized implementating that functionality myself was that I needed more : multi line text,etc...

I think adding functionality is always attractive at first, especially for newcomers but there should be a limit to not bring misleading expectations about what Kha is about.

For that particular issue, I am open and maybe we need the multi line version too. 

After all text i such basic thing to expect from a media framework and not being able to draw it at a specific place without extra code might be a disappointing experience

if so, I am with sh-dave for brining into the g2 extension

FacelessJ:
Seems like the consensus is at the very least the alignment should be in GraphicsExtension rather than G2, so have moved it over.

RobDangerous:
OK, let's do it. GraphicsExtension will probably move out of Kha soon but it'll take the String alignment functions with it to wherever it goes.

Issue #653, opened 2017-09-03
get font.width/height blocks fontGlyphs


RblSb:
Example:
https://gist.github.com/RblSb/75e4aa11fe81b5e38c7d08d523102ae9
Font:
https://fonts.google.com/specimen/Open+Sans
Even after receiving w/h, fontGlyphs can not be set.
Tested on html5/flash/android.
The problem with calling kravur._get(fontSize).getHeight() without the glyphs parameter https://github.com/Kode/Kha/blob/master/Sources/kha/Kravur.hx#L126

Perhaps g.fontGlyphs and the like would be more convenient in the Font class? Something like https://pastebin.com/5avHh8Db

PR #672, opened 2017-10-18
font.width/height fix

RblSb:
Please speak if you know the solution better. Сlosed #653.

### DIFF
--- a/Sources/kha/Kravur.hx
+++ b/Sources/kha/Kravur.hx
@@ -171,12 +171,12 @@ class Kravur implements Resource {
 		return images[fontSize];
 	}
 
-	public function height(fontSize: Int): Float {
-		return _get(fontSize).getHeight();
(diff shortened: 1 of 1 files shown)


RblSb:
@RobDangerous сan I add a method for unloading only font images? This is necessary when changing fontGlyphs, otherwise the font characters will not be recached. Like:
```haxe
public function unloadImages(): Void {
	images = new Map();
}
```
Also, at the moment it's necessary to specify a full array of characters without omissions (0 ... 1000 instead of 0...100, 900...1000). You can generate an array of omissions in __get(), and then use it in getCharWidth(), like:
```haxe
gaps = [];
var next = 0;
for (i in 0...glyphs.length) {
	if (glyphs[i] != next) {
		if (i > 0) gaps.push(glyphs[i-1]);
		gaps.push(glyphs[i]);
		next = glyphs[i] + 1;
	} else next++;
}
```
But then you have to pass this array to getCharWidth()... Maybe you will offer a better solution?

PR #686, opened 2017-11-12
Firefox keyPress fix

RblSb:
(which == 8) is backspace in FF, all other nonchars is 0


RobDangerous:
Henceforth you shall be called the Firefox extinguisher.

PR #689, opened 2017-11-15
g2.fontGlyphs fix (non-static)

RblSb:
Alternative option with optional fontGlyphs argument. Sorry for recreating =) Сlosed #653.
Working example:
https://gist.github.com/RblSb/a0ecc7393b370422f66de25f2cfb2a85


RobDangerous:
Picked #688 instead.

PR #704, opened 2017-12-09
renamed Audio & Video capture to AudioCapture & VideoCapture

lewislepton:
just a simple renaming. keeping the folder layout.


RobDangerous:
I think you also have to change something in the HTML5 backend to make that work.

lewislepton:
ah ok. will do

lewislepton:
forgot to recursive clone so just trying it again & get some tests done. should be done soon

lewislepton:
ok. tested properly after recursive clone. all working

Issue #716, opened 2017-12-26
Latest commit breaks the apps in Google Chrome


riseremi:
**Context**
Target: HTML5
Chrome version: 63.0.3239.84 (NOT the latest)
Kha version: latest from the master (https://github.com/Kode/Kha/commit/1f090a6f6a7b6a78013e60dc57c317176757500d)

**Problem**
I've updated Kha and now I get this error in the console:
```
Uncaught TypeError: Cannot read property 'isVrEnabled` of undefined
```
However, it works in Safari and Firefox. I guess the problem is in Chrome.
If I manually checkout this commit (https://github.com/Kode/Kha/commit/d68a17637a96d3eb47cda16dae63e1120f404e73) it works as well.

![image](https://user-images.githubusercontent.com/6680627/34359219-0acb388e-ea67-11e7-8cff-beb2d4ec279f.png)

![image](https://user-images.githubusercontent.com/6680627/34359221-0d5e996a-ea67-11e7-93f5-e9edd40b9eb9.png)

PR #717, opened 2017-12-26
Add VrInterface.instance null check

riseremi:
Hello. I managed to fix the error mentioned in https://github.com/Kode/Kha/issues/716
Please, let me know if this is ok.


RobDangerous:
Uhm, sorry but I already fixed that a minute ago.

riseremi:
Ok, thanks.

PR #727, opened 2018-01-14
Update Vector2.hx

Triangle345:
Add angle calculation


RblSb:
Maybe `function angle(v: FastVector2): FastFloat`?

Triangle345:
good catch with FastVector2 param.

Triangle345:
thx https://github.com/Kode/Kha/pull/729

RobDangerous:
I think you'll have to create a new pull request.

PR #738, opened 2018-01-24
Keep backward compatible

luboslenco:
I would like to keep Kha backward compatible whenever possible, any objections about this gamepad api? I may have overlooked something, please double check.


juakob:
Looks fine, it could be cleaner to have them both in the same function

juakob:
I mean it's cleaner to have diconnect and connect callbacks in the same function

RobDangerous:
Agreed.

PR #759, opened 2018-02-24
implement changeResolution html5

juakob:
implement changeResolution html5

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -1014,7 +1014,10 @@ class SystemImpl {
 	}
 
 	public static function changeResolution(width: Int, height: Int): Void {
-
+		if (untyped khanvas.getContext) {
+			khanvas.width=width;
+			khanvas.height=height;
+		}
 	}
 
 	public static function setKeepScreenOn(on: Bool): Void {


wighawag:
Does it work ?

the code at https://github.com/Kode/Kha/blob/master/Backends/HTML5/kha/SystemImpl.hx#L445 should prevent that from working.

to change resolution in html5, the canvas can be resized externally (for example using css). This will resize the size of the canvas on screen too though

While there could be reason to want the canvas to be of certain resolution (performance is the one I think of but this should be tested), it will rarely be independent form the actual size of the canvas on screen (clientWidth/clientHeight)

RobDangerous:
I'm still unsure about merging anything to do with resizing the canvas because it all depends on the embedding html/css.

juakob:
I test it in Firefox,Chrome and Edge, they all scale the canvas. I had to do it to reset the scale once you go out of fullscreen, otherwise the canvas keeps the fullscreen size. It does not work in KS, but there you dont have the problem.

I know nothing about html/css, but I think its external to what Kha and the generated js can do. If you embed the js in some way other than the default way, you are going to need a specific solution.

This implementation did the trick for me in the default solution, but maybe its not generic enough, I just made it work for me. I thougth it was a better solution than nothing.

PR #761, opened 2018-03-01
Made HTML5 target respect devicePixelRatio

FuzzyWuzzie:
As per [https://www.khronos.org/webgl/wiki/HandlingHighDPI](https://www.khronos.org/webgl/wiki/HandlingHighDPI), these changes correctly size the `khanvas` on the screen (using css) and on the buffer (using width/height).

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -437,17 +437,17 @@ class SystemImpl {
 			if (untyped canvas.getContext) {
 
 				// Lookup the size the browser is displaying the canvas.
-				//TODO deal with window.devicePixelRatio ?
-				var displayWidth  = canvas.clientWidth;
-				var displayHeight = canvas.clientHeight;
+				if(canvas.clientWidth != options.width ||
+				   canvas.clientHeight != options.height) {
 
-				// Check if the canvas is not the same size.
-				if (canvas.width  != displayWidth ||
-					canvas.height != displayHeight) {
+					// set the display size of the canvas.
+					canvas.style.width = options.width + "px";
+					canvas.style.height = options.height + "px";
 
-					// Make the canvas the same size
-					canvas.width  = displayWidth;
-					canvas.height = displayHeight;
+					// set the size of the drawingBuffer
+					var devicePixelRatio = js.Browser.window.devicePixelRatio == null ? 1 : js.Browser.window.devicePixelRatio;
+					canvas.width = options.width * devicePixelRatio;
+					canvas.height = options.height * devicePixelRatio;
 				}
 
 				System.render(0, frame);


Backends/HTML5/kha/SystemImpl.hx
```diff
@@ -437,17 +437,17 @@ class SystemImpl {
...
-				var displayWidth  = canvas.clientWidth;
-				var displayHeight = canvas.clientHeight;
+				if(canvas.clientWidth != options.width ||
+				   canvas.clientHeight != options.height) {
 
-				// Check if the canvas is not the same size.
-				if (canvas.width  != displayWidth ||
-					canvas.height != displayHeight) {
+					// set the display size of the canvas.
+					canvas.style.width = options.width + "px";
```
wighawag: This is going to break external css styling
on the web, the canvas need to be considered out of control (at least in term of sizing)

RobDangerous: Ya, I wondered, wasn't the previous code at that position problematic already?

wighawag:
This should have been sufficient to deal with dpi, no :
```
  var devicePixelRatio = window.devicePixelRatio || 1;
  var framebufferWidth = canvas.clientWidth * devicePixelRatio;
  var framebufferHeight = canvas.clientHeight * devicePixelRatio;
   if(canvas.width != framebufferWidth || canvas.height != framebufferHeight){
     canvas.width = framebufferWidth;
     canvas.height = framebufferHeight;
  } 
```


But I think we might need to add some option to not always use the full space as if I remember correctly there are limitations on some browser that cannot handle canvas too big

FuzzyWuzzie:
I agree there has to be a better way of dealing with canvas sizing, but until then I figured we could start with code that touched the same bits as what was already there (which did touch CSS and the canvas size).

Perhaps an extra option for `System.init` That allows for more fine-grained control?

wighawag:
That's not true, the previous code was not touching the css. It was reading the display size and setting the canvas size (which is the framebuffer size, not the visual size)  to be equal (ignoring dpi though)

FuzzyWuzzie:
I would propose something like a flag that indicates if the site's CSS should dictate the size of the canvas or if you want Kha to. If Kha, then use the code as merged. If use the site's CSS, then don't touch the canvas style at all and use the alternate code that @wighawag posted. Also perhaps expose a cross-platform parameter for the devicePixelRatio?

wighawag:
I disagree, kha should never touch the css and the solution proposed in the comment above (https://github.com/Kode/Kha/pull/761#issuecomment-369499104) should already solve your dpi issue (assuming you have the css style that set the width and height (instead of setting them up in the options)). 

Then we have to deal with canvas's framebuffer being too big to handle (but this is probably already an issue (it is not dpi specific just more likely to happen when dpi scale is used) on multiple screen setup with a browser window taking all the space. 

We could also make an option to not scale to dpi and keep the canvas framebuffer scaled down (though I am not sure this is ever desired)

In regard to adding more flexibility to Kha for case where canvas's framebuffer need to be fixed (maybe for performance reason or code simplicity) I proposed a solution in the code "inline comment" above (https://github.com/Kode/Kha/pull/761#discussion_r171479366), this should solve the other potential canvas sizing wished by several member while making sure Kha do not interfere with styling so Kha game can be embeded anywhere if wished.
I ll see if can work on it in the coming weeks.

FuzzyWuzzie:
@wighawag you're right about the old code, I made this pull request late last night and must have been seeing things. Your version needs a slight tweak as it infinitely up-sizes the canvas. I can make another pull request to address this, but first Kha's propensity for controlling the size needs to be sorted in regard to the HTML5 target.

Namely: 

* khamake's `project.windowOptions.width` and `projection.windowOptions.height` options (which setup the width & height of the canvas in the HTML, and default to 800x600 if not given)
* `System.init()`'s width & height options (which default to 800x600 if not given)

Do we leave these as is and just ignore them? The issue there is we would then need to include some css on the generated HTML to properly size the canvas, otherwise on a retina display the canvas would display at double the desired size.

So I would propose something along the lines of:

1. Change khamake's html5 template to size the canvas using css instead of `width` & `height`
2. Change the canvas sizing code to resize the framebuffer based on devicePixelRatio
3. Expose the devicePixelRatio in `System` for easier cross-compatibility

Thoughts?

FuzzyWuzzie:
Moving this into new pull requests: Kode/khamake#131 and Kode/Kha#761...

Backends/HTML5/kha/SystemImpl.hx
wighawag: Actually there is no need for kha to support keepRatio if a fized size is chosen since you can use the css ```object-fit``` property : https://drafts.csswg.org/css-images/#the-object-fit that allow various option to deal with difference between display size and canvas framebuffer size

PR #762, opened 2018-03-01
 Made html5 framebuffer use devicePixelRatio and css for sizing

hamaluik:
As per the discussion in Kode/Kha#761, this should properly size the canvas' framebuffer by taking into account the window's `devicePixelRatio` and the size of the canvas on the screen. Note: in order for this to not throw everything into chaos, it should be merged at the same time as Kode/khamake#131 which changes the html5 template to use css styling instead of directly applying the size to the width and height of the canvas element.

If the size of the canvas isn't controlled by css and this code is used, then the canvas will infinitely try to resize itself larger and larger on HighDPI screens (where `pixelDeviceRatio != 1.0`).

This and Kode/khamake#131 should be discussed by at least @RobDangerous and @wighawag before merging.

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -437,17 +437,17 @@ class SystemImpl {
 			if (untyped canvas.getContext) {
 
 				// Lookup the size the browser is displaying the canvas.
-				//TODO deal with window.devicePixelRatio ?
-				var displayWidth  = canvas.clientWidth;
-				var displayHeight = canvas.clientHeight;
+				if(canvas.clientWidth != options.width ||
+				   canvas.clientHeight != options.height) {
 
-				// Check if the canvas is not the same size.
-				if (canvas.width  != displayWidth ||
-					canvas.height != displayHeight) {
+					// set the display size of the canvas.
+					canvas.style.width = options.width + "px";
+					canvas.style.height = options.height + "px";
 
-					// Make the canvas the same size
-					canvas.width  = displayWidth;
-					canvas.height = displayHeight;
+					// set the size of the drawingBuffer
+					var devicePixelRatio = js.Browser.window.devicePixelRatio == null ? 1 : js.Browser.window.devicePixelRatio;
+					canvas.width = options.width * devicePixelRatio;
+					canvas.height = options.height * devicePixelRatio;
 				}
 
 				System.render(0, frame);


wighawag:
That code should not result in infinitely resize of the framebuffer as it does not resize the display size of the canvas, just its framebuffer. Did you experience the chaos you describe ?

The main idea of not manipulating the css is basically to allow the html and css to be anything. As such the previous template code (pre https://github.com/Kode/khamake/pull/131 ) should work fine as while the width and height are set on the canvas at the beginning they are changed back to reflect the display size of the canvas once that code get executed.

I nevertheless think it is a good thing to change the template as you did in https://github.com/Kode/khamake/pull/131 put both of these changes should be work independently

FuzzyWuzzie:
Yea I tested it and unless the css was controlling the size, on a high res display (devicePixelRatio = 2), every frame it tried to double its size (Firefox, OSX). I'm not totally sure why because like you said, it should work. But ¯\_(ツ)_/¯

I totally agree about not manipulating the css, and I also usually use my own "template" (just throw a custom `index.html` into the assets folder), but if this is done then the other change is necessary as far as I can tell.

wighawag:
Oh I see now why it expends,
It actually makes sense because if no css is supplied the browser ensure the display size default to be the size of the framebuffer (and thus dynamically)

It was not an issue without dpi adjustment because both display size and framebuffer size were made equal by kha anyway.

I think we should find a way to avoid this problem.

One quick solution would be to make dpi scale, an non default option. User who want to use dpi scale, would then have to understand the implication of using it (they would need to make sure the canvas has some css applied that set its width and height)

There might be other solution though.
I tried to use ```window.getComputedStyle(canvas,null).getPropertyValue("width")...``` but this return the size even if no css is set

I tried to check ```canvas.style.width``` but this is not the only way to have the canvas width set (a class name can do it for example, or a style on all canvas

So these would not work

I understand that your template fix the issue but the template is used for development. When the user will embed its game in someone else website and this website do not use css to size the canvas, chaos will happen.

Backends/HTML5/kha/SystemImpl.hx
```diff
@@ -436,18 +436,13 @@ class SystemImpl {
...
-					canvas.style.width = options.width + "px";
-					canvas.style.height = options.height + "px";
-
-					// set the size of the drawingBuffer
-					var devicePixelRatio = js.Browser.window.devicePixelRatio == null ? 1 : js.Browser.window.devicePixelRatio;
-					canvas.width = options.width * devicePixelRatio;
-					canvas.height = options.height * devicePixelRatio;
+				// resize the framebuffer based on the devicePixelRatio and the canvas size
+				var devicePixelRatio = Browser.window.devicePixelRatio == null ? 1 : Browser.window.devicePixelRatio;
+				var framebufferWidth = canvas.clientWidth * devicePixelRatio;
```
wighawag: We should probably use Math.floor as devicePixelRatio can be an non integer
But it seems that setting canvas.width/height performs that operation anyway (at least on chrome)

FuzzyWuzzie:
So I've been looking around at other frameworks / engines / whatever, and as far as I can tell they all do one of two things:

1. Completely ignore devicePixelRatio and just set or use the canvas width & height as they're presented to them
2. Use CSS or set the element's style to control the size and use devicePixelRatio

So, as far as I can see, there isn't a way around using styling when using devicePixelRatio at this time. Thus I would agree with your "quick solution": add an option to enable high-dpi scaling (disabled by default), documenting the fact that you'll need to then control the styles of the canvas element.

Thoughts?

wighawag:
I think it make sense. The only reason to use dpi scale is when the canvas'
framebuffer size is depending on the canvas's display size. When the canvas
framebuffer is fixed there is no use for the dpi scale option.

On 6 Mar 2018 8:13 p.m., "Kenton Hamaluik" <notifications@github.com> wrote:

So I've been looking around at other frameworks / engines / whatever, and
as far as I can tell they all do one of two things:

   1. Completely ignore devicePixelRatio and just set or use the canvas
   width & height as they're presented to them
   2. Use CSS or set the element's style to control the size and use
   devicePixelRatio

So, as far as I can see, there isn't a way around using styling when using
devicePixelRatio at this time. Thus I would agree with your "quick
solution": add an option to enable high-dpi scaling (disabled by default),
documenting the fact that you'll need to then control the styles of the
canvas element.

Thoughts?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Kode/Kha/pull/762#issuecomment-370912763>, or mute the
thread

PR #768, opened 2018-03-09
Pen input for Kore

luboslenco:
To go with https://github.com/Kode/Kore/pull/248.

### DIFF
--- a/Backends/Kore/kha/SystemImpl.hx
+++ b/Backends/Kore/kha/SystemImpl.hx
@@ -4,6 +4,7 @@ import kha.input.Gamepad;
 import kha.input.KeyCode;
 import kha.input.Keyboard;
 import kha.input.Mouse;
+import kha.input.Pen;
 import kha.input.Sensor;
 import kha.input.SensorType;
 import kha.input.Surface;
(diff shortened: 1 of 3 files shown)


sh-dave:
Does it make sense to actually implement that via the already existing surface api? Thinking of touchscreens that don't necessarily have a pen.

luboslenco:
I am not sure. The pen input gets registered above the tablet, before it even touches the surface. On top of pressure there is tilt, rotation, additional buttons on the pen. Plus usually we want both pen and touch input, touch for manipulating the canvas and the pen for actual drawing. With that said it could still be mangled into surface class, but this seemed like the cleanest way.

sh-dave:
Oh ok, didn't knew how pen's actually work. The api just looked kinda similar. So yeah separate pen api it is 😅

tienery:
This could also be used in conjunction with drawing tablets connected to a PC, which will use pen input. I'm curious of Kore/Kha would pick up drawing tablets for input, though. Maybe this new pen API could pick that up somehow? Or if it's considered a surface, perhaps `surfaceConnected` and `surfaceDisconnected`.

RobDangerous:
This is for drawing tablets connected to a PC, have a look at the Kore pull request.

tienery:
Does it make sense to use `pressure` for the `penUp` listener? We could remove that as releasing the pen from the device is the removal of any such pressure.

luboslenco:
penUp may be called when the pen stops hovering over the tablet(if supported), not when the tip leaves the surface. While the pen is hovering it can still interact with the tablet. (eg programmable button artifically applying pressure).

PR #772, opened 2018-03-16
Prevent browser keys

### LINKED ISSUE #733: Web page scrolls on pressing cursor keys in HTML5 game
When pressing any navigation key in a HTML5 game, the web page will scroll along. Remark that this is not only for the cursor keys, but also for any other browser navigation. Page up/down, spacebar, .... In Firefox, backspace will navigate history to the previous page, etc.

As I see it, there are 2 possible solutions:

1. Disallow any browser keys and shortcuts (including for example ctrl-t for opening a new tab)
2. Disallow any key that was handled by the game. So it's up to the game to decide if it handled a certain key, and if not it's passed on to the browser.

Remark that the above only disallows keys when the game canvas has focus. When it does not, key events are not passed on it and so navigation works as expected.

The implementation of this functionality can be found in Backends/HTML5/kha/SystemImpl.hx. It contains 3 functions that react to javascript key events: keyDown(), keyUp(), keyPress().

keyDown() contains a call to event.stopPropagation(), but remark that this will not stop the browser from reacting to keys. only event.preventDefault() will make sure the browser does not handle the key.

At first, adding event.preventDefault() to keyDown() might seem like a good solution (keyUp() already has it), but unfortunately this will prevent keyPress() to be triggered, and that one handles kha text input with keyboard.sendPressEvent(char).

Only doing event.preventDefault() for non character keys is not possible because of spacebar.

I would think handling character keys in keyDown(), and always doing event.preventDefault() would fix it. Kha would then ignore keyPress(). But I don't know how keys on mobile (coming from the virtual keyboard) are handled.

RblSb:
Blocks all non-char keys in new browsers, does nothing in the old ones.
keyCode 16/18 is Shift/Alt (do not block shift + char, etc)
Sample: https://gist.github.com/RblSb/84f94798a5bef36d7aa32a0839e40559
Close #733

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -910,6 +910,11 @@ class SystemImpl {
 	}
 
 	private static function keyDown(event: KeyboardEvent): Void {
+		if ((untyped event.code == event.key)
+			|| event.ctrlKey || event.metaKey
+			|| event.keyCode == 16 || event.keyCode == 18) {
+				event.preventDefault();
+		}
 		event.stopPropagation();
 
 		// prevent key repeat
@@ -938,8 +943,8 @@ class SystemImpl {
 	}
 
 	private static function keyPress(event: KeyboardEvent): Void {
+		event.preventDefault();
 		event.stopPropagation();
-		if (firefox && (event.which == 0 || event.which == 8)) return; // Firefox bug 968056
 		keyboard.sendPressEvent(String.fromCharCode(event.which));
 	}
 


RobDangerous:
Oh, let's see what happens...

RblSb:
@RobDangerous I like that you have a spirit of adventure.

RobDangerous:
You need to have that to have any chance of getting ahead of all the browser madness.

jakelewis3d:
Good work here.
I'd suggest adding 'Back Space',  e.keyCode == 8 to the blacklist, as (on Microsoft Edge at least)  the Back Space key is mapped to the browser Back button - so the whole app unloads!
Jake

RblSb:
@jakelewis3d I changed this code here #773 
Please check https://jsfiddle.net/RblSb/fLtqouxe/ in Edge.
(the backspace should not call an alert or exit the page)
if the problem remains, try this one too:
https://jsfiddle.net/RblSb/fLtqouxe/43/

jakelewis3d:
Great. Both of those work in Microsoft Edge to prevent back space from
unloading the page. Thanks, Jake

PR #773, opened 2018-03-17
Fix keys prevention

RblSb:
Found a shorter solution. Fix work in safari (event.key is null), and also works in IE11. Maybe even works in IE9.


RobDangerous:
Excellent!

PR #776, opened 2018-03-19
Compute for Krom

luboslenco:
Krom does compute shaders! To go with https://github.com/Kode/Krom/pull/66.


RobDangerous:
Krom does all the things.

PR #790, opened 2018-03-30
Fix compile

luboslenco:
Some `;` disappeared from the sources, I am not sure if those are all but kore & krom seem to be fine now.

Also, the code style in Kha is becoming fragmented, which hurts quick navigation through the file. I don't care about particular style but some tool to help us keep this unified would be handy.


RobDangerous:
Ah, thanks, saved me the trouble. We have a checkstyle file but so far I couldn't get it to work well enough to actually use it.

sh-dave:
Sry, my fault. I wrapped the call with a try / catch in kore/krom, but then removed them later on as it didn't actually work. forgot to put the `;` back. Had a look at the haxe-checkstyle thing a while back as well, but same result as robert 😞 I do watch the repo and it seems to be in active development, so maybe i'll give it a shot again some time.

PR #801, opened 2018-04-12
[Hashlink] Added the additional parameters for Assets

Jarrio:
I also added the `pos` parameter for consistency ~ did not check if it was required.


sh-dave:
I was a minute faster with #800 👅

Jarrio:
But I added the `pos` position :D

sh-dave:
But it'snot that useful here imo. The point in having it in `kha.Assets.loadXXX()` was to show the file/line where it's called from, and it's only used in the default version that just uses `haxe.Log.trace`.

RobDangerous:
Sorry, Dave won the race.

PR #810, opened 2018-04-23
fix missed Key -> KeyCode changes

sh-dave:
I think this wasn't changed when kha.Key was removed in favor of kha.input.KeyCode?


RobDangerous:
Probably.

PR #828, opened 2018-06-13
Support all formats in WebGLImage.getPixels

luboslenco:


Didiercode:
Seems that the "getPixel" (not getPixels) or "at" not available again in last version ? 

(for a standard usage like here after a 	image = RenderPathCreator.finalTarget.image; when doing something like thereafter
trace('pixel($x,$y) = ${image.at(x,y)} '); or 	
trace('getPixel($x,$y) = ${image.getPixel(y,y)} ');

PR #835, opened 2018-07-05
System.language feature

RblSb:
Implemented for flash, java, js backends. Need some help regarding cpp things.


Sources/kha/System.hx
```diff
@@ -171,6 +171,15 @@ class System {
 		return SystemImpl.getSystemId();
 	}
 
+	/**
+	 * The IS0 639 system current language identifier.
+	 */
+	public static var language(get, null): String;
```
Gama11: Shouldn't this be `(get, never)`? `(get, null)` means "set access only within this class / read-only to the outside", so there's still an underlying field (which doesn't seem needed here).

RblSb: @Gama11 Thanks, fixed =)

Issue #854, opened 2018-08-06
AudioChannels are not advancing without registered `audioCallback`


sh-dave:
To handle muting all sound and music when our app goes into the background, i use `kha.audio2.Audio.audioCallback = null;` This seems to work at first, but exhibits a problem.

Every sound that is played during this `mute period` doesn't advance it's internal state, but will only do it once i set `kha.audio2.Audio.audioCallback` back to it's default mix function.

RobDangerous:
Hm, I didn't consider audioCallback = null to be more than a hack to silence the computer during development, for actual production use we should do something else (like setting the volume to zero and doing an early return in the mixer when there's nothing to mix).

sh-dave:
But it seemed such an easy fix **\*sigh\***

PR #855, opened 2018-08-06
[html5] implement audio muting/unmuting

sh-dave:
Is this anything as you proposed it in #854?

- use global gain nodes and connect children (the sounds to play) to it
- control global gain node via `mute/unmute()` in `kha.audio2.Audio` and `kha.js.MobileWebAudio`

### added public api

```haxe
class kha.audio2.Audio {
    public static function mute() { ... }
    public static function unmute() { ... }
}
```

### todo 
- dunno if or how to handle `AudiolementAudio`, `VirtualAudioElement`
- no public `master-volume` api yet, as i don't need that for now; but i guess it would be easy to implement via the global `GainNode`
- add mute / unmute stubs or implementations for the remaining backends


sh-dave:
@RobDangerous have any opinion on this?

RobDangerous:
Not yet sure, might prefer to implement that directly in the mixer for Audio2.

sh-dave:
i implemented that by just setting channel volumes to 0 in our app now

PR #891, opened 2018-10-02
Clean up

RblSb:
Fix defines, add System.language for WPF


RobDangerous:
Oh, that's nice. I like clean ups.

PR #892, opened 2018-10-02
fontGlyphs refactoring

RblSb:
This variable is static, so it makes no sense to pass it everywhere.
The iteration through the glyphs array is 20x faster in the case of [this test](https://try.haxe.org/#0669b) and should be much better in real code.


RblSb:
Additionally:
- Move charBlocks to g2.Graphics, make it private
- Add setter for fontGlyphs for generating charBlocks after fontGlyphs change, not with first drawString/Font.width/height
- use @:privateAccess for charBlocks in Kravur class
@RobDangerous, thoughts?

RobDangerous:
I don't understand @:privateAccess, goes way over my head.

Issue #847, opened 2018-07-26
Krom's cutCopyPaste


ustance:
Krom do nothing with `System.notifyOnCutCopyPaste`

Tested on this version, using built-in Krom:
![Kode Studio](https://i.imgur.com/WfdoiLL.png)

Here is a simple test. It works in html5-debug, but not in Krom 
![Test](https://i.imgur.com/O26c7l8.png)

RobDangerous:
That's on Windows, right?

ustance:
Yep, Windows 10, x64.

PR #893, opened 2018-10-07
Cut copy paste for Krom

luboslenco:
To go with https://github.com/Kode/Krom/pull/98.

Fixes https://github.com/Kode/Kha/issues/847.


RobDangerous:
API versioning is in, the pull request is merged - but all of that will stay in the krom_api_1 branch until the next Kode Studio release (planned for next next weekend). From then on we can up the API anytime but I had to change Krom a little to wait for a start event so all of that can work in a non-hacky way.

PR #899, opened 2018-10-23
Made WindowFeatures an abstract enum

Jarrio:
Makes the parameter a bit clearer at first glance and provides completion

### DIFF
--- a/Sources/kha/System.hx
+++ b/Sources/kha/System.hx
@@ -1,5 +1,7 @@
 package kha;
 
+import kha.WindowOptions;
+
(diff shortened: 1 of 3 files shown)


RobDangerous:
Uh oh, this doesn't compile.

Jarrio:
How odd, I just tested the code on try haxe and it seems to work over there https://try.haxe.org/#C3D08

RobDangerous:
But that's Haxe 3, Kha is on Haxe 4 preview 5 by default.

PR #901, opened 2018-10-28
Quaternion constructor now inline.

Disar:
Please let me know if `getEulerAngles` and `slerp` should also be inlined.  Ill make a new pr for those.


RobDangerous:
It's ok like that I think.

RobDangerous:
Also Haxe now has https://github.com/HaxeFoundation/haxe/pull/7425 so when people want the big ones to be inlined, they can easily do that.

Disar:
oh that's sexy

PR #904, opened 2018-11-08
support x/y window position in debug-html5

sh-dave:
to go with https://github.com/Kode/khamake/pull/170

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -66,11 +66,10 @@ class SystemImpl {
 		}
 		var wndOpts = {
 			type: 'showWindow', title: options.title,
-			x: options.x, y: options.y,
+			x: options.window.x, y: options.window.y,
 			width: options.width, height: options.height,
 		}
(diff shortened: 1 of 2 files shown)


Backends/HTML5/kha/SystemImpl.hx
```diff
@@ -64,7 +64,12 @@ class SystemImpl {
 		if (electron.webFrame.setZoomLevelLimits != null) { // TODO: Figure out why this check is sometimes required
 			electron.webFrame.setZoomLevelLimits(1, 1);
 		}
-		electron.ipcRenderer.send('asynchronous-message', {type: 'showWindow', title: options.title, width: options.width, height: options.height});
+		var wndOpts = {
+			type: 'showWindow', title: options.title,
+			x: options.x, y: options.y,
+			width: options.width, height: options.height,
```
RobDangerous: Can you use options.window.x here instead of adding an additional SystemOptions parameter?

sh-dave: Sure, done.

PR #905, opened 2018-11-09
instancedRenderingAvailable is true for Krom

luboslenco:
Krom is more capable than advertised.


RobDangerous:
We need a better marketing department.

PR #910, opened 2018-11-23
use enum abstracts

sh-dave:
Damn, so many whitespace changes...

anyways, the enums are now abstracts, and all the `foo.getIndex()` is changed to just `foo` due to automatic conversion to `Int`.

Compiled on html5, krom and native.

### DIFF
--- a/Sources/kha/graphics4/Usage.hx
+++ b/Sources/kha/graphics4/Usage.hx
@@ -1,6 +1,6 @@
 package kha.graphics4;
 
-enum abstract Usage(Int) to Int {
+@:enum abstract Usage(Int) to Int {
 	var StaticUsage = 0;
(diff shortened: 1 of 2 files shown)


Sources/kha/graphics4/Usage.hx
```diff
@@ -1,7 +1,7 @@
 package kha.graphics4;
 
-enum Usage {
-	StaticUsage;
-	DynamicUsage; // Just calling it Dynamic causes problems in C++
-	ReadableUsage;
+enum abstract Usage(Int) to Int {
```
RblSb: There is some `enum abstract` instead of `@:enum abstract`

RobDangerous: Aaaaaaah

PR #912, opened 2018-11-23
Added KHA and HXCPP defines for Kore builds trough Kha

Disar:


RobDangerous:
I'm reverting the HXCPP define because as far as I can tell it's not defined by regular hxcpp and it's tested for in hxcpp in Xml.cpp and if it is defined that results in a compile error - that one is easy to fix but I want to stay as close to original hxcpp as possible to minimize merge trouble.

PR #917, opened 2018-11-28
Add setTextureCompareMode

luboslenco:
This lets us (after all these years) sample the depth textures / shadow maps faster.

After enabling the compare mode using `g.setTextureCompareMode(unit, true);`, `sampler2DShadow` & `samplerCubeShadow` will work like expected. krafix already has the support built-in.

I kept the implementation as simple as possible for now, nothing fancy. Works for Krom, hl/c, hxcpp, html5 + opengl, webgl2, d3d11.

To go with https://github.com/Kode/Kore/pull/329 and https://github.com/Kode/Krom/pull/109.

### DIFF
--- a/Backends/Android/kha/android/Graphics.hx
+++ b/Backends/Android/kha/android/Graphics.hx
@@ -277,6 +277,14 @@ class Graphics implements kha.graphics4.Graphics {
 	
 	}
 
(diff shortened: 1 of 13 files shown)


RobDangerous:
Yay!

PR #933, opened 2018-12-19
Minor dce improvements

### LINKED ISSUE #923: @:keep minimizatiton
With an @:keep class, we save all its functions and functions that are called from its functions. Maybe we should use @:keep only for individual fields if possible?

https://github.com/Kode/Kha/blob/master/Sources/kha/Assets.hx
I don't see an opportunity to get an error because of the absence of this class. `Reflect.field(Assets.images, "img")` will not work anyway without assets loading.

https://github.com/Kode/Kha/blob/master/Sources/kha/internal/BytesBlob.hx
Keeped because of [this line](https://github.com/Kode/Kha/blob/master/Backends/Kore/kha/Storage.hx#L23). We can add `Blob.alloc(0);` to this function instead?

Reason for Sensor (and other input classes) keeping?
https://github.com/Kode/Kha/blob/master/Backends/HTML5/kha/input/Sensor.hx
https://github.com/Kode/Kha/blob/master/Backends/Kore/kha/input/Sensor.hx
https://github.com/Kode/Kha/blob/master/Backends/KoreHL/kha/input/Sensor.hx

And maybe with haxe `@:ifFeature("kha.package.*")` we can get rid of ogg decoder automatically, without #kha_no_ogg? [How it works](http://try-haxe.mrcdk.com/#572CC)

RblSb:
Closes #923

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -408,7 +408,6 @@ class SystemImpl {
 		}
 		else {
 			SystemImpl._hasWebAudio = false;
-			AudioElementAudio._compile();
 			untyped __js__ ("kha_audio2_Audio1 = kha_js_AudioElementAudio");
 		}
 
--- a/Backends/HTML5/kha/input/Sensor.hx
+++ b/Backends/HTML5/kha/input/Sensor.hx
@@ -2,7 +2,6 @@ package kha.input;
 
 import kha.SystemImpl;
 
-@:keep
 class Sensor {
 	private static var isInited: Bool = false;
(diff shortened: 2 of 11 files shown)


RobDangerous:
Uh oh, those look dangerous. Those @keeps were added by others for reasons I think.

RblSb:
@RobDangerous i made some test projects for Sensor, Font and Assets with/without loadEverything and Reflect.field, have no problems on kore targets. You can recheck that there is no cpp @:functionCode links to kha classes for Assets, Blob, Sensor and Font, then it should be safe, i guess.

RobDangerous:
Did some git blame... @juakob are you ok with removing the Sensor @keep?
Also, I want my inlines back.

PR #954, opened 2019-01-12
Use Kore::FileReader for HL

luboslenco:
Makes HL take advantage of the `Kore::FileReader`, ie. reading blobs works perfectly on Android now.

Once `Kore::FileReader` is ported to C, we can patch the `hl_file_contents()` directly at:
https://github.com/Kode/Kha/blob/master/Backends/KoreHL/hl/src/std/file.c#L146

Like we do for hxcpp at:
https://github.com/Kode/khacpp/blob/cc0097dd4822f75c707b6e87e6acfec68bcf38d1/src/hx/libs/std/File.cpp#L347

(HL + Android build times are awesome!)


RobDangerous:
Cool Spot!

PR #958, opened 2019-01-17
Image.getPixels() for non-render-targets

luboslenco:
Right now `Image.getPixels()` fails for non-render-target images. This allows to retrieve bytes for `readable` images. Started with Krom.

- `Image.getPixels()` - read pixels
- `Image.lock()/unlock()` - modify pixels

Depends on https://github.com/Kode/Kore/pull/351 and https://github.com/Kode/Krom/pull/118.


vivichrist:
YAY!!

PR #986, opened 2019-02-11
Khabind: webidl binder

zicklag:
**Discussion:** https://github.com/armory3d/haxebullet/issues/25

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -30,3 +30,7 @@
 	path = Kinc
 	url = https://github.com/Kode/Kinc.git
 	branch = master
+[submodule "Tools/webidl"]
+	path = Tools/webidl
+	url = https://github.com/ncannasse/webidl.git
+	branch = master
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -60,38 +60,40 @@ class SystemImpl {
 	}
 
 	public static function init(options: SystemOptions, callback: Window -> Void): Void {
-		SystemImpl.options = options;
-		#if kha_debug_html5
-		Browser.window.onerror = cast errorHandler;
-		var electron = untyped __js__("require('electron')");
-		if (electron.webFrame.setZoomLevelLimits != null) { // TODO: Figure out why this check is sometimes required
-			electron.webFrame.setZoomLevelLimits(1, 1);
-		}
-		var wndOpts = {
-			type: 'showWindow', title: options.title,
-			x: options.window.x, y: options.window.y,
-			width: options.width, height: options.height,
-		}
-		electron.ipcRenderer.send('asynchronous-message', wndOpts);		// Wait a second so the debugger can attach
-		Browser.window.setTimeout(function () {
+		kha.Assets.loadKhabindJsLibs(() -> {
+			SystemImpl.options = options;
+			#if kha_debug_html5
+			Browser.window.onerror = cast errorHandler;
+			var electron = untyped __js__("require('electron')");
+			if (electron.webFrame.setZoomLevelLimits != null) { // TODO: Figure out why this check is sometimes required
+				electron.webFrame.setZoomLevelLimits(1, 1);
+			}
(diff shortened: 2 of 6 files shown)


RobDangerous:
This isn't needed anymore, right?

zicklag:
As far as I understand there isn't anything that made it unneeded. I recently made sure that it was all up-to-date and working ( see https://github.com/armory3d/armory/pull/1159#issuecomment-517879675 ).

In Armory there were some physics glitches that were present in the HashLink builds that have now been ported to the Krom builds because my PR makes the Krom builds use the same bindings as the HashLink builds, but those were just glitches that needed to be fixed anyway because they were already present in HashLink before my PRs. Those are just Armory problems, though.

The binder itself is working perfectly as far as I can tell and there aren't any problems that I know of that would prevent merging it to Kha.

sh-dave:
can't you just put the code generation into one of the khamake callbacks and load the libs manually 
 before `System.init` or bundle via `--macro includeFile("something.js")`?

zicklag:
Maybe you could do something like that but there is a lot of logic in the code that went into the process of using the Haxe WebIDL binding library to create the bindings as C++ files, calling Emscripten to compile and link the source files, and only if they have been changed, etc..

I think it would probably be a small to medium chunk more work and configuration if you wanted to do it manually. The whole point of the PR was to make bindings work automatically for both HashLink, Krom, and the Browser with a minimal amount of configuration or manual intervention.

Being that the work of automating it is already done with these PR's I don't know why you would rather do it manually unless there is something wrong with the implementation.

sh-dave:
well b/c it's code someone has to maintain eventually. you can still keep most of the code and just move it to some utility library that does the heavy lifting i guess. and i'm not saying there is something wrong, i personally just don't think it actually belongs in the kha codebase.

zicklag:
That's a reasonable point. I don't know exactly what it would look like to move it to a separated module, but if that was possible it could be a better way to do it.

RobDangerous:
The main point here is, I don't think you will use it anymore since you're doing your own engine now and so far nobody else asked for it so I suspect this will just be a bunch of code and even an additional dependency which nobody will use.

zicklag:
As shown https://github.com/armory3d/armory/pull/1159, other community members are asking about it, not me. The point is not that I would use it myself, but that it would be a benefit to Armory and the Armory community.

RobDangerous:
I see one person asking about a bullet update in Armory. That seems to be only vaguely related to this pull request and I don't even know whether Lubos wants to go this route to achieve it.

luboslenco:
Sorry, if core Kha contributors have concerns then I don't feel I should force it.

The less code there is to maintain in Kha the better (and this would fall onto me now as you have your own engine) - exploring the way Dave suggested would be the best option.

https://github.com/armory3d/armory/pull/1159#issuecomment-531591916

Amir-Arsalan:
@RobDangerous Just letting you know that this is actually something that many people have asked for in many different ways on Armory forums.

luboslenco:
To be clear this has nothing to do with physics interactions in Armory. This is related to how we internally bridge C/C++ code to run in Kha/Armory.

RobDangerous:
A search for webidl in the Armory forums only shows zicklag himself talking about it. Please don't muddle the discussion with vague statements.

zicklag:
My thought was that this would help *reduce* code maintenance in Armory by automating the bindings and removing the `#if js ... #else ... #end` code throughout the Armory physics implementation. If it would be more of a burden to maintain the binder than it would be to maintain the manual bindings then that is your call.

RobDangerous:
I'll close it then. Lubos, if you ever want something like this, let's go the Dave route. Personally I'm not interested in automatic bindings and would strongly prefer not having to maintain complicated functionality related to it.
Sorry it didn't work out, zicklag.

Amir-Arsalan:
@zicklag @luboslenco Wouldn't it possible to have a forked version of this repo with these changes that comes with Armory so that we don't have to rely on the current official release of this repo?

PR #1003, opened 2019-03-03
Fix Image.at for WebGL images created via Image.fromBytes

Joncom:
When an `Image` has been created via `Image.fromBytes` (WebGL), its `image` property is a `Uint8Array` (has no `width` or `height`), but the broken code accesses these non-existent properties.

Broken:
`var r = data.data[y * Std.int(image.width) * 4 + x * 4];`

Fixed:
`var r = data.data[y * image.width * 4 + x * 4];`

Explanation:
`Std.int(image.width)`
becomes:
`undefined | 0`
becomes:
`0`

Therefore, `Image.at` multiples `y` with `0`, returning an incorrect result for all `y > 0`.

### DIFF
--- a/Backends/HTML5/kha/WebGLImage.hx
+++ b/Backends/HTML5/kha/WebGLImage.hx
@@ -123,10 +123,10 @@ class WebGLImage extends Image {
 			else createImageData();
 		}
 		
-		var r = data.data[y * Std.int(image.width) * 4 + x * 4];
-		var g = data.data[y * Std.int(image.width) * 4 + x * 4 + 1];
-		var b = data.data[y * Std.int(image.width) * 4 + x * 4 + 2];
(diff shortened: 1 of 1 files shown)


Joncom:
You can where the `kha.WebGLImage.image` property is set to a byte array that has no `width` or `height`, here: https://github.com/Kode/Kha/blob/ebafec87c6422e949489931d2493edf84947f814/Backends/HTML5/kha/Image.hx#L48

Joncom:
Oops, made a typo in the comment explaining this pull-request.

The fix is actually:
`var r = data.data[y * width * 4 + x * 4];`
not
`var r = data.data[y * image.width * 4 + x * 4];`

The fix is already correct in the actual code though, so all is well in that regard.

PR #1010, opened 2019-03-10
Fix CompilerDefines fields

RblSb:


RobDangerous:
That's a very "for my use only" pull request :-P
Let's replace '-' with '_' and prepend '_' to keywords.

RblSb:
This is more `for hxcpp use` pr :)
Replacement need field checking for duplicates. Will see how can i check added fields or use StringMap. But you probably don't need if (CompilerDefines._static != null), when you can use #if, same for other bool defines, doesn't? This is why pr is fine.

RblSb:
https://github.com/HaxeFoundation/haxe/issues/5002 is fixed, so Kha probably broken with current behavior, so i complete this thing.

RobDangerous:
I fixed it so much, it will never ever be a problem again (also added an updates.md and referenced this pull request in there).

RblSb:
I mostly interested in this class for window name instead of default:
https://github.com/Kode/Kha/blob/master/Sources/kha/System.hx#L7
So you don't need to set same project name in `System.start`, only in `khafile.js`. Can we move `Macros.hx` to `kha.internal` and set `title = Macros.projectName()` as default value?

RobDangerous:
I don't think the Window title should be a define, just pass it manually.

RblSb:
That's sad, then this line should be removed https://github.com/Kode/khamake/blob/master/src/main.ts#L120

RobDangerous:
But using that line you can write some code which sets the window title to the project name.

RblSb:
But you say that this should not be a define :) If you fine with define in khamake then i'm good, but i can also add this define manually in khafile, without khafile dependency.

RobDangerous:
It's already there and today I don't feel like removing more things.

Issue #1006, opened 2019-03-06
[Request] Expose underlying Graphics4 to Graphics2 when g2 is on g4


hamaluik:
To do almost anything with a custom `pipeline` on a `Graphics2` instance, you need access to the underlying `Graphics4` implementation so you can do things like set extra textures, etc.

As far as I can tell, this isn't made available anywhere. I have a kludgy hack that only barely works right now (I find the juggling of Graphics / Graphics2 / Graphics2 on Graphics4 / Backend Graphics / etc very confusing, so if I can sort it out I'll open a pull request).

Thanks!

PR #1014, opened 2019-03-21
Added getter for g4 when g2 is on g4

hamaluik:
When g2 is running on an underlying g4 instance, we can set a custom pipeline. However if that pipeline deviates at all from the default pipelines, there aren't really any ways to control it. Implementation for #1006

For example, in my use-case, my custom pipeline samples from two textures (the default one provided by ImageShaderPainter, and a custom one in a custom shader I'm using). In order to set that second texture, I need access to the underlying g4 instance, hence this small change to give access.

Tested and works in my use-case, but let me know if there are any problems.

### DIFF
--- a/Sources/kha/graphics2/Graphics.hx
+++ b/Sources/kha/graphics2/Graphics.hx
@@ -228,6 +228,12 @@ class Graphics {
 		setPipeline(pipeline);
 		return pipe = pipeline;
 	}
+
+	public var g4(get, never): kha.graphics4.Graphics;
+	
+	private function get_g4(): kha.graphics4.Graphics {
+		return null;
+	}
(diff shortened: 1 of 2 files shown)


hamaluik:
@RobDangerous are there any issues with this?

RobDangerous:
Haven't made my mind up on whether it makes sense yet because you can just get g4 from the surface you got g2 from. It's convenient to have this of course but on the other hand I try to keep things minimal.

hamaluik:
Oh man, I feel like a dunce now. Of course you can! It's not quite as intuitive as this I think, but I agree with you about keeping things minimal. I'm happy to close this without merging if you are.

PR #1015, opened 2019-03-25
Allow html color strings for Color.hx

Jarrio:


RobDangerous:
Uhm, I don't think I want that. Any compelling reasons you think that this is worthwhile?

sh-dave:
Creating colors from strings should be the exceptional case if you care about performance. With this change it might happen by accident and you won't even notice it.

Jarrio:
@RobDangerous Most IDE integrated color tools work with string values, so it just adds a simple visual cue for what color you're using. 
image
Could I ask why you're opposed? It doesn't affect performance and is purely a QoL change

@sh-dave What performance issues are there? This is an abstract, it doesn't exist during runtime, there would be zero performance impact using `0x0000ff` vs `#0000ff`

RobDangerous:
fromString is not an abstract an will indeed be slower than using 0x0000ff directly - considerably so in fact.

sh-dave:
There will be quite some performance issue, as all the string parsing code (the whole body of the function) would be used.

Jarrio:
Ah, it appears like that was my own ignorance - my bad! I still personally like the idea of having the option between either side as it does have some utility just not out in the exported code, I guess.

RobDangerous:
Just do `this.graphics.color = Color.fromString("0x0000ff")`

Jarrio:
That's what I was originally doing, I had just thought with this being an abstract this was actually an oversight and not an intention. 

I corrected some of my knowledge on abstracts anyway so it looks like I got PR for my brain, so it's still a win 👍

RobDangerous:
There's some confusing messaging in the Haxe community about Haxe abstracts being a compile time construct - that's only true in so far as abstracts do not introduce any new runtime types or allocations but the methods declared in an abstract end up being regular functions in the resulting code, there's no compile-time aspect to them.

Jarrio:
I think it stems from the natural learning progression, I had initially started to use abstracts as just `@:enum abstracts` where this is actually true and as you get more and more ingrained in the ecosystem I think you just kind of "assume" that it also applies to everything else "abstract"

PR #1016, opened 2019-03-27
Bitmap baking optimization (over 100x speedup)

FaB0SS:
"font._get(30).getTexture();" with default glyphs, takes:
Before optimization:  ~10000 ms
After oftimization: ~85 ms


RobDangerous:
Haha, you're the BOSS!

PR #1017, opened 2019-03-27
Fix surface move event

RblSb:
Oh god...


RobDangerous:
Interesting pull requests coming in today...

PR #1037, opened 2019-04-19
Add Graphics2.scale function

Joncom:


Sources/kha/graphics2/Graphics.hx
```diff
@@ -128,6 +128,10 @@ class Graphics {
 
 	public var transformation(get, set): FastMatrix3; // works on the top of the transformation stack
 
+	public function scale(x :FastFloat, y :FastFloat): Void {
+		transformation = kha.math.FastMatrix3.scale(x, y).multmat(transformation);
```
sh-dave: Does this inline properly? Otherwise `transformation.setFrom()` should be used i think.

ibilon: Lots of functions from this file could be made inline too.

PR #1041, opened 2019-04-26
Optimize android-java

RblSb:
3x faster on bunnymark


RblSb:
Now 6x faster, but still bad in comparison. Someone want to test jni ways like this?
https://www.badlogicgames.com/wordpress/?p=904

PR #1043, opened 2019-05-06
KravurImage minor fix

RblSb:
Use `Image.fromBytes` if it's implemented.


RobDangerous:
What does it fix?

RblSb:
`Image.fromBytes` is faster on some targets

RobDangerous:
Did you profile that? Looking at the Kore implementation, fromBytes actually looks worse. Think I'll have to have a closer look at those functions.

RblSb:
Well, only with `haxe.Timer.measure` x)

RobDangerous:
And what targets? And what are the results?

RblSb:
Kore android and webgl, seems always faster. Webgl:
Current: 0.01
fromBytes: 0.0009
Kore:
Current: 0.15
fromBytes: 0.01
With any order.

RobDangerous:
That's a lot. But we should have a look at that readable parameter in Kore's fromBytes, that makes it worse in other ways. And let's add fromBytes everywhere so we can get rid of the double whopper implementation before merging this.

RblSb:
`fromBytes` is not implemented only on flash/android-java, will check it later. So, we should add `readable:Bool = false` as `fromBytes` last argument?

RobDangerous:
Yes, I think we should do that.

RblSb:
Achtung: this patch can cause crashes for some androids in `Kore::Graphics4::Texture::init(char const*, bool)`. Can't get more details sadly. Ideas?

RblSb:
Lets revert KravurImage change and merge only api change?

PR #1057, opened 2019-06-04
prevent duplicate channels

sh-dave:
## steps to reproduce

- play sounds: `sound0`, `sound1`, `sound2`
- shortly after, pause `sound2`
- wait until `sound0` and `sound1` finished
- resume `sound2`
  - it will get inserted in `soundChannels[0]` (as `sound0` is already finished) and therefore appear twice in the array
  - this will in turn call `AudioChannel.nextSamples` twice per tick and speedup the playback


RblSb:
It makes sense to remove `soundChannels[i] == channel` on next line `if` then?

PR #1070, opened 2019-06-26
fix a lot of haxe4 rc3 warnings

sh-dave:


RblSb:
If Robert still wants haxe3 compat, we probably can add some typedefs to haxe package. But i think nobody uses haxe3 flag.

RobDangerous:
At this point not worth it anymore me thinks.

PR #1073, opened 2019-06-27
AssetData type

RblSb:
It's time to properly type our code! https://github.com/HaxeFoundation/haxe/issues/8484

### DIFF
--- a/Sources/kha/Assets.hx
+++ b/Sources/kha/Assets.hx
@@ -5,6 +5,30 @@ import haxe.Unserializer;
 
 using StringTools;
 
+typedef AssetData = {
+	name: String,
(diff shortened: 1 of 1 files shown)


RblSb:
Meh, let's skip this i guess

sh-dave:
why did you close this?

RblSb:
Robert says this typedef is hard and restrict custom file types.

RobDangerous:
"Robert says" is reason enough for anything and everything.

sh-dave:
hm ok then

(but technically, it doesn't restrict anything as you can still Reflect.field() on the object, you just won't see it in code completion)

PR #1079, opened 2019-07-11
Reduce fileLoad functions

RblSb:
Before:
```haxe
imageLoad: function(done,failure) {
	var tmp;
	if(failure != null) {
		tmp = failure;
	} else {
		var f = haxe_Log.trace;
		var infos = { fileName : "kha/internal/AssetsBuilder.hx", lineNumber : 132,
			className : "kha._Assets.ImageList", methodName : "imageLoad"};
		tmp = function(v) {
			f(v,infos);
		};
	}
	kha_Assets.loadImage("image",function(image) {
		done();
	},tmp,{ fileName : "kha/internal/AssetsBuilder.hx", lineNumber : 132,
		className : "kha._Assets.ImageList", methodName : "imageLoad"});
}
```
After:
```haxe
imageLoad: function(done,failure) {
	kha_Assets.loadImage("image",function(image) {
		done();
	},failure,{ fileName : "kha/internal/AssetsBuilder.hx", lineNumber : 132,
		className : "kha._Assets.ImageList", methodName : "imageLoad"});
}
```

### DIFF
--- a/Sources/kha/Assets.hx
+++ b/Sources/kha/Assets.hx
@@ -321,6 +321,6 @@ class Assets {
 		return LoaderImpl.getVideoFormats();
 	}
 
-	public static inline function reporter(custom: AssetError -> Void, ?pos: haxe.PosInfos)
+	public static function reporter(custom: AssetError -> Void, ?pos: haxe.PosInfos)
 		return custom != null ? custom : haxe.Log.trace.bind(_, pos);
 }
--- a/Sources/kha/internal/AssetsBuilder.hx
+++ b/Sources/kha/internal/AssetsBuilder.hx
@@ -166,7 +166,7 @@ class AssetsBuilder {
 						}, {
 							value: null,
 							type: Context.toComplexType(Context.getType("kha.internal.AssetErrorCallback")),
-							opt: null,
+							opt: true,
 							name: "failure"
 						}]
 					}),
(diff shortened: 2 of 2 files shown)


sh-dave:
Looks like the default `trace` won't be called anymore in case you don't provide a custom reporter function?

RblSb:
Error with pos info should be triggered inside of loadImage/loadEtc ([there is reporter call](https://github.com/Kode/Kha/blob/master/Sources/kha/Assets.hx#L233)).
I checked CORS error (from file://) and get errors like this in both cases:
`kha/Assets.hx:174: {url : sounds/fx_powerup_use_repair.mp3}`
So i think it should work for other cases too.

RblSb:
How to trigger `AssetsBuilder.hx` error? Maybe that `posInfo` not required? Still think this is good change for dce and code size with many assets.

sh-dave:
Oh, i actually only use the `loadXXXFromPath` functions so i don't have anything like that generated. But yeah i guess in your case it would save quite some size for these functions. To trigger an error it should be possible to stop after `System.start` and just delete a file out of the `build/target` folder.

RblSb:
Tested `loadImageFromPath`, `loadEverything` and `imageLoad`, they all gives error to `Assets.hx` with `url: image.png` link. I only get `AssetsBuilder.hx` error with `Assets.images.imageLoad(()->{}, null)`. Should `error` arg be optional?

RblSb:
Should be fine now.

RblSb:
Maybe we can simply remove these macro functions and add `Assets.load(r:Resource)/unload` or `resource.load()/unload()`?

PR #1089, opened 2019-08-02
Use Haxe 4 new function type syntax with argument names for Mouse.notify/remove

nadako:
It's much more clear and also IDEs can generate listeners with proper argument names.


RobDangerous:
Well, technically we're still compatible to Haxe 3 via the --haxe3 option but I think I broke that a while ago. So, let's merge.

PR #1090, opened 2019-08-03
rebuild Shaders module whenever manifest file is changed

nadako:
This fixes completion/compilation-with-cache when adding/removing shaders.


nadako:
sorry for the whitespace changes, my editor trims them and yours should too!

nadako:
Also you know what would be cool? If `files.json` contained the path to the original glsl file, then we could use that for the `Shaders.shader_name` field position, so "go to definition" would open the glsl file! \o/

RobDangerous:
The files.json glsl needs a GitHub issue by you.

PR #1092, opened 2019-08-04
fix canvas.getContext check

nadako:
With that `untyped`, Haxe will generate `if($bind(canvas,canvas.getContext))` which doesn't make sense, however with the simple null check it's going to be `if(canvas.getContext != null)` which I believe was the intention here.

Related: https://github.com/HaxeFoundation/haxe/issues/6297


nadako:
notice me senpai!

PR #1093, opened 2019-08-07
fix -dce full for native builds

sh-dave:


RobDangerous:
I award you the title of Dungeon @Keeper.

PR #1101, opened 2019-08-11
Keyboard.disableSystemInterventions

### LINKED ISSUE #1064: Keyboard.disableSystemInterventions
Enum with `Full / None / Custom(keys:Array<KeyCode>)` constructors to change default key behavior in html5.

RblSb:
I added `Default` for current option (allow F-keys and chars, blocks ctrl/cmd hotkeys and backspace).
Also i think `Custom(func: (code: KeyCode)->Bool)` would be more useful than `Array<KeyCode>`.
Thoughts?
Closes #1064

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -980,7 +980,7 @@ class SystemImpl {
 			case Full:
 				event.preventDefault();
 			case Custom(func):
-				if (func(event.keyCode)) event.preventDefault();
+				if (func(cast event.keyCode)) event.preventDefault();
 			case None:
 		}
 		event.stopPropagation();
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -14,6 +14,7 @@ import js.html.DeviceOrientationEvent;
 import kha.graphics4.TextureFormat;
 import kha.input.Gamepad;
 import kha.input.Keyboard;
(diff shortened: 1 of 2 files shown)


RblSb:
Found that firefox clipboard hack is not requred, `canvas.onpaste` works, the problem was in `onkeydown preventDefault` for x/c/v.
Good time to start adding browser-specific keycodes and unify them. How about this:
```haxe
case 186: Semicolon;
case 187: Equals;
case 189: HyphenMinus;
```
The problem that this is breaking change. Only firefox uses original `kha.input.KeyCode` values for these methods.
[from this](https://github.com/ccampbell/mousetrap/pull/215):
```
  ;   FF  59  Chrome 186  IE 186  Saf 186  
  =   FF  61  Chrome 187  IE 187  Saf 187 
  -   FF 173  Chrome 189  IE 189  Saf 189 
```
So then `if (key == 186)` in user code cannot be catched. We can change `Semicolon/etc` values and break `if (key == Equals)` instead.

RblSb:
Robert, do you want to change these keycodes to chrome version?
Also would be interesting to find list of keycodes for native targets.

RobDangerous:
The Kinc keycodes are here: https://github.com/Kode/Kinc/blob/master/Sources/kinc/input/keyboard.h
Ideally that should always be used.

RblSb:
Okay, will change Kha's `KeyCode` class about these three keys to Kinc version and will do browser-specific checks to make it work.

RobDangerous:
Super cool, thanks!

PR #1109, opened 2019-08-17
Implement Storage for HL

RblSb:


RobDangerous:
Hey, we talked about this - this pull request is not correct, the Kinc API can't be used that way and what you did only works because the Kinc-side non-console implementation is not yet correct. Sorry.

RblSb:
Should i use Kore's FileReader instead?

RobDangerous:
Yes. I'll change the Storage interface when I get to it though.

RblSb:
Ok, will wait for async interface first.

RobDangerous:
Okidoki. Sorry again, please send letters to Nintendo, Sony and Microsoft about how annoying this is.

RblSb:
Thats fine, just played a little around c, hl and kinc, even fix fun problem. Hlc generates pretty clean code btw :p

RobDangerous:
Yes and it compiles fast. But it runs slower than hxcpp and as long as that's the case I'll prioritize hxcpp.

PR #1111, opened 2019-08-18
Normalize White Space (Unity Backend)

williamthompsonj:
All line endings now \n so unity stops throwing warnings about line endings

Forced single new line at end of file

Removed white space in following places:
- On right side of line
- Between code and semi-colon
- Empty brackets, braces, and parens
- Beginning of file
- Turn 3+ new lines into 2


RobDangerous:
Aaah, why \r\n again? Does Unity require that?

williamthompsonj:
No reason I suppose, I haven’t checked the rest of the code base to see what the most normal line ending currently is.

RobDangerous:
I don't know what it currently is either but it should definitely be \n.

williamthompsonj:
I’m good with that, I’ll resubmit with just \n line endings and default to that for anything else.

RblSb:
For some reason first commit here has normal diffs, when second with correct line endings has full file diffs. This is strange...

RblSb:
Yes, Kha just uses `\r\n` for most files. Oops. Lets merge all PRs and fix that after =)

RobDangerous:
The aftermath of using flashdevelop so many years ago.

PR #1120, opened 2019-09-05
Vertexbuffer/Indexbuffer unlock count

RudenkoArts:
Added for HL/Web target vertexbuffer unlock count, and for Web indexbuffer unlock count


RudenkoArts:
I don't know if we need to have indexbuffer.unlock(count). :|

RobDangerous:
I sure need it.

PR #1121, opened 2019-09-11
More enum abstracts

RblSb:
Functions with int switches removed. `@:expose` from `kha.Color` too.

### DIFF
--- a/Sources/kha/compute/Access.hx
+++ b/Sources/kha/compute/Access.hx
@@ -1,11 +1,7 @@
 package kha.compute;
 
-enum abstract Access(Int) {
+enum abstract Access(Int) to Int {
 	var Read = 0;
 	var Write = 1;
 	var ReadWrite = 2;
-
-	public inline function toInt():Int {
-		return this;
-	}
 }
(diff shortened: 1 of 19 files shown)


Sources/kha/compute/Access.hx
```diff
@@ -1,7 +1,11 @@
 package kha.compute;
 
-enum Access {
-	Read;
-	Write;
-	ReadWrite;
+enum abstract Access(Int) {
```
sh-dave: just add `to Int` and remove the toInt() function (if you need the int conversion at all)?

RblSb: I think explicit `to` functions is little better for such enums (when you need to get int only for internal apis), but `to Int` is fine too

sh-dave: just wondering, why do you think it's better?

RblSb: You can send that abstract as int argument accidentally and switch only for `Read` and `Write` without errors, but this is very minor =)

PR #1123, opened 2019-09-15
Minimize AssetsBuilder

RblSb:
Removes `resLoad, resUnload, resName` and `names` array, because they can be accessed with `Type.getInstanceFields` and `resDescription`. Less allocated strings and less generated code.

### DIFF
--- a/Sources/kha/Assets.hx
+++ b/Sources/kha/Assets.hx
@@ -8,9 +8,7 @@ using StringTools;
 @:build(kha.internal.AssetsBuilder.build("image"))
 @:keep
 private class ImageList {
-	public function new() {
-
-	}
+	public function new() {}
 
 	public function get(name: String): Image {
 		return Reflect.field(this, name);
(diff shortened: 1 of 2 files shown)


RobDangerous:
Those are convenience functions, I don't want to remove them. Sorry.

RblSb:
Eh, it's just saves 700kb of generated cpp code for 4,9mb project size. Some flag for these fields maybe, just to have descriptions and file fields? Also my `loadEverything` looks slightly nicer =)

RobDangerous:
Can DCE not help here? And yes, loadEverything looking different was another problem with the pull request, you mixed a clean-up and a change in one pull-request.

RblSb:
Hm, DCE can help in theory, i need to mark file fields and Descriptions with `@:keep` meta and remove `@:keep` from Lists classes. But reflection for "res" + "Unload" start being a problem, but i guess this is not why these functions exist. Are you interested in separated `loadEverything` change?

PR #1124, opened 2019-09-15
Align System.language

RblSb:
Return primary code on all targets.


RobDangerous:
Awesome pull request!

PR #1126, opened 2019-09-21
Loader Dce

RblSb:
Another try with same optimization. `loadEverything` depends only on `Description` files, `Load/Unload/Name/names` still available with direct field access.

### DIFF
--- a/Sources/kha/Assets.hx
+++ b/Sources/kha/Assets.hx
@@ -6,11 +6,8 @@ import haxe.Unserializer;
 using StringTools;
 
 @:build(kha.internal.AssetsBuilder.build("image"))
-@:keep
 private class ImageList {
-	public function new() {
-
-	}
+	public function new() {}
 
 	public function get(name: String): Image {
 		return Reflect.field(this, name);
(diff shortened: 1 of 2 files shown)


RobDangerous:
What does this change? It's too hard to see for me.

RblSb:
`Load/Unload/Name/names` fields are removed from assets lists if not used in code.

RobDangerous:
I don't understand how that happens. Can you split this up in a purely functional pull request without the clean ups? Githubs change view makes a mess of things as it is.

RblSb:
Actual change is removed `@:keep` from `ImageList/SoundList/etc` and added `@:keep` only to generated `Description` and `file` fields (`meta: [{pos: pos, name: ":keep"}]`), so by default `fileUnload` and other stuff not generated, if you don't call it as `Assets.images.fileUnload()` or `Assets.images.fileName;`.
There is also dependent `Assets.loadEverything` change to use only `Description` files for her mission, without using `fileName/fileLoad` fields for correct dce removings.

RobDangerous:
For reference, this was unmerged immediately after it was merged and then replaced by #1128

RblSb:
Dark magic happen

PR #1128, opened 2019-09-25
Simplify loadEverything

RblSb:
`loadEverything` only uses `Description` files (calls local `loadFunc` for file types instead of macro generated `Load` functions), so dce (or define) can remove other fields from AssetsLists in next pr step. Not sure if this helps, better to read new version instead of diff...


RobDangerous:
So this is now on part of the Loader DCE pull request which means I'm supposed to merge this and close the other one, right?

RblSb:
Right

RobDangerous:
Nooo, now I did it the other way round.

RobDangerous:
The force push is with me.

Issue #1099, opened 2019-08-10
kha.Image.fromCanvas for HTML5


nadako:
pretty please

RobDangerous:
Hey, it's fixed.

PR #1134, opened 2019-10-02
Added Image.fromCanvas for HTML5 target

hamaluik:
As requested in Kode/Kha#1099

Tested and confirmed working in Firefox for canvas, webgl1, and webgl2 backends.


RobDangerous:
Cool Spot!

PR #1151, opened 2019-11-05
New typed arrays structure and classes

matrefeytontias:
Establishes a documentable structure for typed arrays. In short :
- ByteBuffer holds the actual data in the form of a fixed-size array of unformatted bytes. No access to the data can be made through this class.
- ByteArray provides mixed-type access to an underlying ByteBuffer, with full control over number type and endianness.
- All typed array types like Uint8Array or Float32Array are abstracts over ByteArray with adequate array access overloads, and all unify with ByteArray and vice-versa.

A pretty sizeable pro of this new structure is that adding a new backend only requires implementing ByteBuffer and ByteArray, as opposed to every single typed array type before. ByteBuffer and ByteArray are declared in the source as extern classes so documentation is generated and a consistent interface is accessible.

Comes with a bonus HTML5 backend implementation with ArrayBuffer and DataView, no extra fees.


matrefeytontias:
Bump. How close is this to being merged ?

RobDangerous:
Not sure but I'll try to get the necessary thinking and speed-testing done this month.

Sanva:
Hi! Any advance on this?

I'm trying to improve the performance of an integration with a 3rd party library that gives back IBOs and VBOs to render them — and this looks nice to avoid some unnecessary type conversions of large buffers.

RobDangerous:
Oh, I should really get to this.

RobDangerous:
Finally checked it. And I like it! In it goes...

matrefeytontias:
Wow, never give up the hope huh 😁

Issue #1147, opened 2019-10-26
Http interface is inconsistent across backends


Huyston:
Http interface of html5 and android for instance is inconsistent.
html5 expects Map<String, String> for the headers and android expects String for the contentType.

This should be very simple to fix tho. Seems like the android backend is lagging behind.

PR #1152, opened 2019-11-08
Implemented request for the Android Backend

Huyston:
This PR implements a simple Http request method for the Android Backend and solves issue #1147.


RobDangerous:
haxe.Http was very restrictive last time I tried (only get and post requests I think). But as discussed in the chat, eventually I'll just move that to user code for non-html5 targets.

Issue #1157, opened 2019-11-21
Window.get_mode Always Returns WindowMode.Windowed on HTML5?


potatoelixir:
Current Behaviour (Backends/HTML5/kha/Window.hx)
```haxe
function get_mode(): WindowMode {
	return Windowed;
}
```

Expected Behaviour
```haxe
function get_mode(): WindowMode {
	return isFullscreen()? Fullscreen:Windowed;
}
```

Is this intentional? Why? What? Currently, to toggle fullscreen on HTML5, I have to use code like this:
```haxe
#if kha_html5
if (@:privateAccess Game.window.isFullscreen())
	Game.window.mode = WindowMode.Windowed;
else
	Game.window.mode = WindowMode.Fullscreen;
#else
if (Game.window.mode == WindowMode.Windowed)
	Game.window.mode = WindowMode.Fullscreen;
else
	Game.window.mode = WindowMode.Windowed;
#end
```
Which is not exactly elegant :P

RobDangerous:
If you know how to fix it, why did you create an issue instead of a pull request? Is it intentional? That's not exactly elegant.

potatoelixir:
1. I don't have any experience with PR's
2. I asked if this is intentional. Someone must've written that code at some point, yes? So for all I know, changing that could break something else in the codebase.

But ok, I'll go ahead and try my best to create a PR and try to become an elegant human being.

PR #1158, opened 2019-11-21
Fix Window.mode getter on HTML5 backend

potatoelixir:
Fixes https://github.com/Kode/Kha/issues/1157 or at least should fix it, untested because I am not able to clone the entire Kha repo for testing right now.


RobDangerous:
Thanks!

Issue #940, opened 2018-12-30
Red and blue color channels are swapped when using graphics1 in Krom.


bent95:
The red and blue color channels are swapped when using graphics1 in Krom. Code that demonstrates the problem:
```
    static function render(frames: Array<Framebuffer>): Void {

        frames[0].g1.begin();
        for (x in 0...frames[0].width){
            for (y in 0...frames[0].height){
                frames[0].g1.setPixel(x,y,Color.Red);
            }
        }
        frames[0].g1.end();
    }
```
This should fill the screen with red pixels, but instead fills it with blue pixels when debugging with Krom. The code works as expected with the HTML5 and Windows targets. This is only a problem with graphics1, similar code using graphics2 does display the correct color.

RblSb:
I think it's because g1 uses RGB colors, while kha.Color stores the ARGB palette.

bent95:
But why does it only happen when debugging in Krom, and not when debugging in HTML5 or running a Windows build?

RobDangerous:
What target system are we talking about?

bent95:
I’m using the latest version of Kode studio running on a 64-bit Windows 7 computer.

potatoelixir:
I know this thread is quite old but... There is already a conditional compilation if/else that swaps `Color`'s  R and B values on HTML5, shouldn't this be as simple as extending that to Krom? I'll submit a PR, please somebody review.

PR #1161, opened 2019-11-23
Fix Graphics1 on Krom

solarane:
Fixes https://github.com/Kode/Kha/issues/940 by having `Graphics1` swap `Color`'s R and B values on Krom like it already does on HTML5, tested using the MWE provided in the issue:
```haxe
fb.g1.begin();
for (x in 0...fb.width) {
	for (y in 0...fb.height) {
		fb.g1.setPixel(x, y, kha.Color.Red);
	}
}
fb.g1.end();
```

### DIFF
--- a/Sources/kha/graphics2/Graphics1.hx
+++ b/Sources/kha/graphics2/Graphics1.hx
@@ -31,7 +31,7 @@ class Graphics1 implements kha.graphics1.Graphics {
 	}
 
 	public function setPixel(x: Int, y: Int, color: Color): Void {
-		#if kha_html5
+		#if (kha_html5 || kha_krom)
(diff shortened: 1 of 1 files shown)


RobDangerous:
What system(s) did you test?

solarrabbit:
I actually forgot what system(s) I made this fix on, because it's been a while... I'll do some more testing and post results.

BlackGoku36:
| Color used to draw | Color actually displayed in krom|
|---------------------|--------------------------------|
|red                           | blue|
|blue                         | red|
|yellow                      | cyan|
|green                       | green |
|Orange                    | Sky blue|
|etc.                        |etc.|

Same sample code from here: https://github.com/Kode/Kha/issues/1184
`img.g1.setPixel(Std.int(mouse.x+y), Std.int(mouse.y+x), kha.Color.Red);`

RobDangerous:
Did some testing and it looks like it's actually swapped everywhere - let's merge.

Issue #1159, opened 2019-11-22
Rendering area has offset on top when device has soft keys


Huyston:
There are some devices that uses virtual buttons (they draw the buttons on the screen, instead of using physical buttons).

For those devices, the returned touch position is summed with the virtual button height, causing a mismatch between the finger position and touch coordinate.

Disabling the virtual buttons fixes the touch position.

Disclaimer: Tested only on Android devices (LG G3 and Samsung S8).

RobDangerous:
Can you fix it? I don't have a device like that.

Huyston:
I'm trying to figure it out.

Huyston:
I think the issue is that we call MotionEvent getX instead of getRawX. getX can return relative positions.

Source: https://developer.android.com/reference/android/view/MotionEvent

RobDangerous:
OK, please give it a try.

Huyston:
Changing that doesn't fix.

potatoelixir:
I have multiple devices I can test this on; I'll see if I can test/check the issue out. Assuming no progress has been made by OP since last comment?

Huyston:
Didn't figure it out yet. Tested on Xiaomi Redmi Note 6, which has soft buttons, but the issue is not present.

Huyston:
There is a black area on top of the devices with this problem, which corresponds with the soft keys height.

System.windowHeight returns a value which is not the actual display height. The viewport 0,0 coordinate is not at the right position. Thus, the coordinate of the touch and the drawing mismatch.

So the problem is not actually the touch coordinate which is wrong, is that the drawing coordinates are wrong for those devices.

Huyston:
Issue is reproducible on the emulator, using Google Pixel 2 API 29 Android 10.

potatoelixir:
Possibly related:
https://stackoverflow.com/questions/29398929/how-get-height-of-the-status-bar-and-soft-key-buttons-bar/29400243
https://stackoverflow.com/questions/6936986/wrong-dimensions-returned-from-display-and-displaymetrics-in-android-3-2/8622079

potatoelixir:
`System.windowHeight` is calculated by code in line 74 of https://github.com/Kode/Kha/blob/master/Backends/Android/kha/Window.hx if I'm not mistaken, and it looks correct D: DisplayMetrics are supposed to take into account status bars, soft menus, etc.

Huyston:
[...]

PR #1163, opened 2019-11-25
Fixes opengl surface height that causes touch coordinates mismatch

Huyston:
The wrong value of height can cause touch coordinate mismatch and prevent Kha from drawing on the entire screen.

This fixes issue #1159.


RblSb:
Add those extern methods here: https://github.com/Kode/Kha/blob/master/Backends/Android/android/opengl/GLES20.hx

Huyston:
Woops, I forgot that if I cange my repo it updates here. It should have been other PR, but I fixed it anyway.

PR #1166, opened 2019-11-26
Fix js_classic blob loading

RblSb:
I don't know bug reason, this flag only removes global closure.

### DIFF
--- a/Backends/HTML5/kha/LoaderImpl.hx
+++ b/Backends/HTML5/kha/LoaderImpl.hx
@@ -182,7 +182,12 @@ class LoaderImpl {
 			loadRemote(desc, done, failed);
 		}
 		else {
+			#if js_classic
+			// readFile never called with remote.require
+			var fs = untyped __js__("require('fs')");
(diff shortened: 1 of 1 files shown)


RobDangerous:
W00t, I don't understand a thing.

RblSb:
If you compile debug-html5 target with `js_classic`, `loadEverything` and at least one blob in your assets - there is stuck on this line https://github.com/Kode/Kha/blob/master/Backends/HTML5/kha/LoaderImpl.hx#L189 (callback is never called, electron 4.0.1).

RblSb:
This thing fixed in electron 6. I recommend to close this PR after vscode-electron-debug update.

RobDangerous:
So the right thing to do is to ignore this pull request and just update Electron instead, right?

RblSb:
Yes, there is khamake electron fix for newest versions.

RblSb:
Fixed

RobDangerous:
Haha

PR #1167, opened 2019-11-26
fix HTML5 crossOrigin issue on iOS/macOS

AlexRamallo:
Apparently Safari on iOS and macOS don't like it when you set the `src` before the `crossOrigin` property. This causes this error:

```
SecurityError: The operation is insecure.
```

This PR fixes that.

Tested on iOS 13.1 and macOS Mojave 10.14

Relevant stack overflow post: https://stackoverflow.com/a/52420950


AlexRamallo:
To clarify: Trying to load a remote image with a URL using `kha.Assets.loadImageFromPath` causes the above error on iOS and Safari, and prevents the image from being loaded. If you disable CORS security features in the developer menu, it will work.

PR #1194, opened 2020-04-25
Fixes playing `.ogg` sound in Android.

Sanva:
This whole thing is explained in https://github.com/armory3d/armory/issues/1412 , and was worked out by @onelsonic and @DarthVernus — I'm simply composing this pull request using their shared code, a bit simplified because it looks like some of the changes are already present.

It just happened to me that my test app was crashing when trying to play `.ogg` sounds in Android, and with these changes it just works fine.

### DIFF
--- a/Backends/KoreHL/kha/korehl/Sound.hx
+++ b/Backends/KoreHL/kha/korehl/Sound.hx
@@ -8,25 +8,18 @@ using StringTools;
 @:keep
 class Sound extends kha.Sound {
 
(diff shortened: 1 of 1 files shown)


RobDangerous:
I'm sorry but that fix is wrong, the different behaviour with wavs and oggs is intentional because uncompressing oggs is very expensive. Applications have to check whether uncompressedData or compressedData is set and act accordingly. CompressedData can be used directly via Audio.stream.

PR #1198, opened 2020-04-27
Implement notifyOnResize for the HTML5 backend

maxfish:
Implementing the support for the window's resize notification.
I'm mimicking the Kore backend's implementation and, for consistency sake, I'm also adding the window 'num', but I'm not sure if it is needed at all in this case.

I've tested the implementation on a HaxeUI project that targets Kha-HTML5

### DIFF
--- a/Backends/HTML5/kha/Window.hx
+++ b/Backends/HTML5/kha/Window.hx
@@ -2,17 +2,26 @@ package kha;
 
 class Window {
 	static var windows: Array<Window> = [];
+	static var resizeCallbacks: Array<Array<Int->Int->Void>> = [];
+	var num: Int;
 	var canvas: js.html.CanvasElement;
 	var defaultWidth: Int;
 	var defaultHeight: Int;
 
 	@:noCompletion
 	@:noDoc
-	public function new(defaultWidth: Int, defaultHeight: Int, canvas: js.html.CanvasElement) {
+	public function new(num:Int, defaultWidth: Int, defaultHeight: Int, canvas: js.html.CanvasElement) {
+		this.num = num;
 		this.canvas = canvas;
 		this.defaultWidth = defaultWidth;
 		this.defaultHeight = defaultHeight;
+		resizeCallbacks[num] = [];
 		windows.push(this);
+
+		var browserWindow = js.Browser.window;
+		browserWindow.onresize = function() {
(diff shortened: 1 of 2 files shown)


Backends/HTML5/kha/Window.hx
```diff
@@ -2,17 +2,26 @@ package kha;
...
+	public function new(num:Int, defaultWidth: Int, defaultHeight: Int, canvas: js.html.CanvasElement) {
+		this.num = num;
 		this.canvas = canvas;
 		this.defaultWidth = defaultWidth;
 		this.defaultHeight = defaultHeight;
+		resizeCallbacks[num] = [];
 		windows.push(this);
+
+		var browserWindow = js.Browser.window;
+		browserWindow.onresize = function() {
```
ianharrigan, line 22: I wonder if, instead of `window.onresize` it should be `window.addEventListener`? The problem is that onresize is a single event listener and therefore be overwritten by other code running on the client (kha could also then be overwriting other important listeners that may be using this single event listener)

ianharrigan:
Looks pretty promising :)

Backends/HTML5/kha/Window.hx
maxfish: Good point, my JS is rusty...
I've changed the code as you suggested.

RobDangerous:
I don't think calling the callback on window resize makes much sense for most people, we should do it on canvas resize.

maxfish:
I understand your point @RobDangerous . I guess the canvas, in the case of the HTML5 target, is the equivalent of the window on native platforms.
As far as I know there is no such event triggered by an HTML component and the implementation would be messy, and I don't think this would make the HTML5 target behave in the same way as, for example, Kore.
But you know the overall architecture, and if you think this should live on the Canvas rather than on the Window, then I'm fine with discarding this PR

RobDangerous:
Yes, it really should. But I'll take this and rework it, no worries.

RblSb:
You can compare canvas element size at the end of `requestAnimationFrame` i think.

mundusnine:
Rob would this method be more suitable as it uses the canvas directly ? : 
image

RobDangerous:
Oh, you're going for the weird and fancy stuff. Yes, should be suitable.

RobDangerous:
@mundusnine did the canvas thing, let's hope it works.

PR #1206, opened 2020-05-14
More compilation fixes for node target

MoritzBrueckner:
When compiling to the node target, I got the following error:

```
.../Kha/Sources/kha/internal/ShadersBuilder.hx:66: characters 50-55 : Array<kha.Blob> should be kha.Blob
.../Kha/Sources/kha/internal/ShadersBuilder.hx:66: characters 50-55 : For function argument 'source'
.../Kha/Sources/kha/Shaders.hx:5: lines 5-7 : Defined in this class
```

The function signatures in the no-backend shader definitions were not the same as the signatures used by the individual backends. Hopefully this does not break any other target (only tested with node, html5 and krom and it worked so far), so lets see if the github build checks will succeed^^

Btw, there should probably be a node build check too :)

### DIFF
--- a/Sources/kha/compute/Shader.hx
+++ b/Sources/kha/compute/Shader.hx
@@ -3,7 +3,7 @@ package kha.compute;
 import kha.Blob;
 
 extern class Shader {
-	public function new(source: Blob, file: String);
+	public function new(sources: Array<Blob>, files: Array<String>);
 	public function delete(): Void;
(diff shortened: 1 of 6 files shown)


MoritzBrueckner:
Hey @RobDangerous, is there anything wrong with this PR? Something I can try to fix?

RobDangerous:
All fine, just busy times.

PR #1215, opened 2020-06-19
add parentheses  to avoid optimization error

juakob:
when use with analyzer-optimize the line gets optimize wrong, even in haxe 4.1

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -173,7 +173,7 @@ class SystemImpl {
 	}
 
 	public static function getTime(): Float {
-		var performance = (Syntax.code("window.performance ? window.performance : window.Date"));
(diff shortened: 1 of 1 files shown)


sh-dave:
Like i wrote in the IRC, i don't see the need for Syntax.code() here.

```haxe
public static function getTime(): Float {
	final now = js.Browser.window.performance != null ? js.Browser.window.performance.now() : js.lib.Date.now();
	return now / 1000;
}
```

RobDangerous:
And didn't you say that it works in Haxe 4.1 yesterday?

juakob:
So sorry Dave, I miss your comment in the log, thats a better solution.
Sorry Rob, I fail you, I thought it was working but later I found that notepad++ was showing the old file build using parentheses

RobDangerous:
ok, then let's use Dave's version.

PR #1219, opened 2020-07-03
Porting ttc support to StbTrueType from stb_truetype.h v1.24

shiena:
For example, Noto Sans CJK's ttc font file can be supported.
see also https://www.google.com/get/noto/help/cjk/


RobDangerous:
Wow, that's a big pull-request. And looking good from what I can make out. Thank you very much!

PR #1227, opened 2020-07-30
[HTML5] Add support for mouse side buttons (back and forward)

zshoals:
Even though it's undocumented, MouseEvent.which does actually support Mouse 4 (back sidebutton) and Mouse 5 (forwards sidebutton).

This has limited use for actual browsers since these buttons are usually assigned to move back and forth through page history and are surprisingly opinionated about these inputs, but works fine on Electron where it's much more useful for debugging purposes.

Considering the inconsistent behavior, I have no issues if you deny the request. Thanks.


RobDangerous:
I like it.

PR #1238, opened 2020-09-03
[hl] Added dynamic comparison support for `hl.F32` and `hl.F64` [fixes #1237]

### LINKED ISSUE #1237: Comparing FastFloats and floats on static targets with dynamic types doesn't work
Hi Robert,

there is a bug with comparing FastFloats and regular floats on static targets if they are stored in `Dynamic` type variables. I'm not sure if it can be fixed easily or if it should be handled by the programmer (= avoiding the Dynamic type in those cases).

**Minimal example code, tested on Krom and Windows C target**:
```haxe
var a: kha.FastFloat = 4.123;
var b: Float = 4.120;

// Without those two lines, both C and Krom targets would print 'true'
var aD: Dynamic = a;
var bD: Dynamic = b;

// Krom: true, C: false
trace(aD >= bD);
```

For more details see https://github.com/armory3d/armory/issues/1814. It doesn't look like a Windows issue, it happens on the native Android target as well.

Also, casting FastFloat values back to floats does not always work on static targets, but I wasn't able to find a minimal reproducible example yet. The error message is `Unsupported type value Single` when compiling .hx code like `v1 = cast(v1, Float);` (v1 is a dynamic variable that holds a FastFloat).

Thanks!

Sanva:
#1237 is indeed a Haxe bug, but in the Hashlink side, part of KoreHL.

### DIFF
--- a/Backends/KoreHL/hl/src/std/cast.c
+++ b/Backends/KoreHL/hl/src/std/cast.c
@@ -353,6 +353,10 @@ HL_PRIM int hl_dyn_compare( vdynamic *a, vdynamic *b ) {
 		return dcompare(a->v.d,(double)b->v.i);
 	case TK2(HI32, HF64):
 		return dcompare((double)a->v.i,b->v.d);
+	case TK2(HF64, HF32):
+		return dcompare(a->v.d,(double)b->v.f);
+	case TK2(HF32, HF64):
+		return dcompare((double)a->v.f,b->v.d);
 	case TK2(HOBJ,HOBJ):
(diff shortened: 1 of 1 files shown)


RobDangerous:
That whole thing wasn't updated for almost a year, we should probably do a complete merge.

Sanva:
@RobDangerous this change is already merged → https://github.com/HaxeFoundation/hashlink/pull/412 .

RobDangerous:
Ah, you did that. In that case I'll merge it so you get extra-credit and then merge the whole thing afterwards.

PR #1247, opened 2020-09-12
Add Uint8array

MoritzBrueckner:
Requires https://github.com/Kode/khacpp/pull/11

On the HL target I've used `unsigned char` instead of `uint8_t` just like the other array types use `short` etc. Is there a reason for that? The byte sizes in the current implementations don't seem to be guaranteed because of that.

I could not test it on Android and Flash but it closely follows the current implementations of the other array types.


RobDangerous:
In hxcpp you mean. No reason, we should change that.

PR #1252, opened 2020-10-02
Added on finished callback for audio channels.

mundusnine:


RobDangerous:
Uh oh, you are calling the callback in the audio-thread, that's going to cause havoc.

RobDangerous:
Had a closer look at the whole situation and it's not worth it to put in some thread-safe callback-mechanism when there's the alternative to just check the finished-property - which is a much easier and more efficient way of doing things in a multithreaded situation. Sorry. But on the positive side, that made me also have a closer look at the current code and the AudioChannel-properties are now truly thread-safe.

PR #1259, opened 2020-10-26
Add VertexData.ColorVertex

RudenkoArts:
targets: Android, Flash, HTML5, Kore, KoreHL


RudenkoArts:
looks like I misunderstood something :D

PR #1267, opened 2020-11-28
[HTML5] Trigger failure callback when a sound file cannot be found

zshoals:
The other guy wasn't going to do it, I think.


RobDangerous:
Haha, I suppose so. Thanks!

PR #1268, opened 2020-11-30
Fix KoreHL Image format

RudenkoArts:
To prevent this type of mistakes i set props setter from null to never 😬

### DIFF
--- a/Backends/Android/kha/Image.hx
+++ b/Backends/Android/kha/Image.hx
@@ -258,7 +258,7 @@ class Image implements Canvas implements Resource {
 		return null;
 	}
 
-	public var g1(get, null): kha.graphics1.Graphics;
+	public var g1(get, never): kha.graphics1.Graphics;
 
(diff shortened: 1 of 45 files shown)


RobDangerous:
What is "this type of mistakes"?

RudenkoArts:
I mean, i accidently set value to image.format, and its property with setter null. Instead of image.myFormat. [here](https://github.com/Kode/Kha/pull/1268/commits/e422dc2b3b84cf53676eb292cbb29cc263d21a61#diff-e2776c67871e33e51e744fe08b5d74c66eb19c1dc806644faee7c60cc3aaf74fL145)

RudenkoArts:
In this [PR](https://github.com/Kode/Kha/pull/1263) i made mistake

nadako:
Note that `get,null` will create a physical field on the class as well, which is completely unused in this case, so `get,never` is much better for getter-only properties.

RobDangerous:
OK, I get it :)

Issue #1277, opened 2020-12-15
Scheduler/Timetasks stop updating while actual framerate significantly higher than detected refresh rate


zshoals:
This describes the actual issue more accurately than https://github.com/Kode/Kha/issues/1266.

**Describe the bug**
It's a little more nuanced than the title.

Scheduler will miss updating Timetasks properly if the actual framerate exceeds Kha's internal detected frequency, causing noticeable hitching and stuttering. The greater the difference between the actual framerate and the detected refresh rate, the more severe the issue becomes. With a big enough difference, updates will stop entirely, unless something interferes with the framerate.

When I say stop, I do actually mean that Timetasks cease execution, or at least the ones with intervals set do.

The issue probably arises from this piece of code: 
https://github.com/Kode/Kha/blob/1891555b39557e74da61d6648fd9087c184dc919/Sources/kha/Scheduler.hx#L193-L198

As an example, when the monitor's refresh rate is 120hz, and the detected frequency is 30hz, this math equates to (0.00833 / 0.033333) ~ 0.25, which is then rounded down to 0, then 0 propagates through the rest of this section. I'm uncertain of the exact behavior afterwards but I assume that Timetasks compare how much of this adjusted time has passed to their target update time, and since the time passed is always 0 updates effectively cease.

I'm not sure if this is a bug with improperly detected refresh rates internally (HTML5 being the big culprit here with its locked 60hz value), or if it's actually a Scheduler issue. Either way the effect it has is pretty...show stopping, you could say.

**To Reproduce**
Steps to reproduce the behavior (probably easiest to test on the HTML5 backend):
1a. Make a simple project that does something noticeable (trace is good enough) in the update loop, or...
1b. ...clone this repo for a simple visual demo: https://github.com/zshoals/timetask-bug
2. Set get_frequency in Display.hx to a low value, let's say 10, which should be low enough to trigger it on any monitor
3. Run it. Notice that updates don't actually occur after the framerate stabilizes after startup.

**Expected behavior**
Dunno exactly, Scheduler is still elusive to me. However, I would assume that Timetasks stopping entirely is not appropriate behavior.

**Execution Environment:**
- Host system (where you compile your code): Windows 7
- Target system (where you run your code): HTML5 (could happen anywhere though, I think)
- IDE used (if any): VSCode
- Kha revision: 1891555b
[...]

PR #1279, opened 2020-12-22
[HTML5] Attempt to estimate device refresh rate at startup

zshoals:
I'm nooby so maybe this isn't a viable approach (or perhaps it's too naive), but I figured I would try adding this feature rather than not try. Sorry if it doesn't work out.

HTML5 does not have a way to just get the device's refresh rate through some query. It can only be estimated at this time.

This code briefly collects the frametime results for successive requestAnimationFrame calls, and tries its best to make an estimate of the monitor's refresh rate before animate() begins and before the System.start callback executes. This test is run at this point so that (hopefully) nothing else important or computation intensive is happening and we get mostly accurate results.

# Caveats:
* Introduces a variable amount of startup delay, dependent on the current refresh rate. I chose 90 frametime samples as a compromise between accuracy and duration of the test since it seemed reasonable (I considered 3 seconds on 30hz as the worst case scenario, 1.5 seconds on 60hz seems passable, etc). I know this isn't great but I don't see a way around it.
* Haven't tested it on toaster ovens. RequestAnimationFrame results there may be more prone to variance caused by computations in the background, which might cause inaccurate results. This may be an issue just generally speaking, but I haven't yet gotten an incorrect refresh rate from my limited tests, however the system I test on is pretty decent.
* Only set at startup once as of now. The refresh rate might possibly change, but I don't think there's a reliable/simple way to know if it actually changed or if something really intensive is occurring due to user code.

### DIFF
--- a/Backends/HTML5/kha/Display.hx
+++ b/Backends/HTML5/kha/Display.hx
@@ -62,7 +62,7 @@ class Display {
 	public var frequency(get, never): Int;
 
 	function get_frequency(): Int {
-		return 60;
+		return SystemImpl.estimatedRefreshRate;
 	}
 
 	public var pixelsPerInch(get, never): Int;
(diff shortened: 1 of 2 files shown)


RobDangerous:
Oooh, super cool! Will first fix #1277 so this doesn't confuse me, then merge your pull request, then add the option to run the test while the application is already running (should work fine while on a title screen for example).

PR #1280, opened 2020-12-23
Implement custom cursor on html5

mundusnine:


lublak:
Custom cursor can be implemented simple by hide the cursor with hideSystemCursor: http://api.kha.tech/kha/input/Mouse.html#hideSystemCursor
And draw yourself the Cursor via drawImage

Also if it implemented this way we should used Assets

RobDangerous:
But that doesn't help - the system-cursor-API is all about using the system's hardware cursor sprite where such a thing is available and thus sidestepping the double-buffered framebuffer and minimize lag in drawing the cursor. You don't get that when using drawImage.

lublak:
Currently for a workaround.

But Assets should be used anyways I think. Or not?

Assets.cursor could be possible. Because on html5 we don't need to load the image to a binary

RobDangerous:
But that workaround gets us nothing. What should actually be used I don't know yet, still gathering information about what's possible on what platform.

lublak:
This workaround is only for now, because nothing ist implemented.
But also as workaround for a single target is to write native code.

My intention was with Assets.cursor that any target can implement his own class Cursor.hx and also the setSystemCursor.
So on html5 we can internal use a relativ path to the image in the assets folder.
On other targets can it be implemented differently.

I only think it's a bad idea to implement it only for html target and without usage of the kha environment (like asset building, etc.)

Sry my English is bad :/

RobDangerous:
Well, I already made Mr Mundus switch his approach.

RobDangerous:
Hm, needs some more work so we don't do all that pixel work again and again when cursors are switched around but it's a start.

mundusnine:
> Custom cursor can be implemented simple by hide the cursor with hideSystemCursor: http://api.kha.tech/kha/input/Mouse.html#hideSystemCursor
> And draw yourself the Cursor via drawImage
> 
> Also if it implemented this way we should used Assets

The reason I made this is to avoid the performance issues Rob talked about when using your workaround because I already do that and it indeed is not performant.

RobDangerous:
drawImage is not performant? Well, that's strange.

RobDangerous:
Just for reference, Mr Mundus found his performance problem and it was indeed unrelated to drawImage.

PR #1303, opened 2021-03-10
Fix segmentation fault in Kinc hxcpp backend when cutting/copying big strings

Sanva:
This is the companion of https://github.com/Kode/Kinc/pull/603 — I've only tested the bug and the solution in Linux, but judging by the code this side of the problem should be affecting other targets, I suppose all that use Kinc hxcpp backend.

### DIFF
--- a/.github/workflows/linux-opengl.yml
+++ b/.github/workflows/linux-opengl.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Apt Update
       run: sudo apt-get update
     - name: Apt Install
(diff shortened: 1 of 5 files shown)


Sanva:
@RobDangerous just a doubt I have while investigating a possible reimplementation for https://github.com/Kode/Kinc/pull/603 — would using `malloc` be acceptable in Kha, taking into account that is is intended to be used with Haxe, a GC language?

RobDangerous:
Yes, in Kha it's ok.

Sanva:
I'm closing this for now, since it looks like I'm accidentally making changes to this pull-request by updating my fork of Kha... sorry for the noise.

PR #1304, opened 2021-03-10
Report HTML5 file drop events

tong:
Not sure why this is hidden behind the `kha_debug_html5`.

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -465,23 +465,18 @@ class SystemImpl {
 		canvas.addEventListener("touchmove", touchMove, false);
 		canvas.addEventListener("touchcancel", touchCancel, false);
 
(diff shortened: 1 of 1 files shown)


tong:
Well, there seems to be no way to  actually read the contents of the dropped file, maybe that's why.
Also wondering how this could fit into current api.

zshoals:
> 
> 
> git blame to the rescue - hey @zshoals, what do you think of this?

git blame lies. This was the first PR I made to Kha and you can see that that conditional compilation block is still in there: https://github.com/Kode/Kha/pull/1227

RobDangerous:
Ah, indeed. It was @sh-dave! Hey Dave, can we remove that #if?

sh-dave:
I think i never tested this in a browser, so sure, go ahead and remove the ifdef if it's working properly.

RobDangerous:
Let's go and hope for the best.

PR #1313, opened 2021-03-23
Mark input listeners as nullable

tyrondis:
Even though `Keyboard.notify()` and `Keyboard.remove()` treat all listeners as nullable, they are not marked as such. This leads to compile errors when using `--macro nullSafety('[...]', StrictThreaded)`. This PR marks the parameters as nullable.


tyrondis:
I just realised that the same problem occurs in other input classes (e.g. `Mouse`). I am going to adjust them as well in a bit.

PR #1314, opened 2021-03-24
Quick solution for  an Iron error.

Lykdraft:
A quick solution to this problem:
https://github.com/armory3d/iron/issues/119

Pretty sure this could be solved in a better way?
Best,
tL.

### DIFF
--- a/Backends/HTML5/kha/graphics4/VertexBuffer.hx
+++ b/Backends/HTML5/kha/graphics4/VertexBuffer.hx
@@ -1,5 +1,6 @@
 package kha.graphics4;
 
+import kha.arrays.Int16Array;
 import kha.arrays.Float32Array;
 import js.html.webgl.GL;
 import kha.arrays.ByteArray;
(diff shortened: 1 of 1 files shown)


sh-dave:
Why don't you just use [armory3d's Kha fork](https://github.com/armory3d/Kha), that already includes this?

Lykdraft:
I do. But git recursive remote. 
So it looks like that the Armory/Kha fork is **behind** the actual Kha fork .
But this PR is not needed anymore anyway I think.

> Hi, thanks for report.
> 
> It looks like `lockInt16()` is no longer available in the kha/html5 backend:
> https://github.com/Kode/Kha/blob/master/Backends/HTML5/kha/graphics4/VertexBuffer.hx
> 
> But it's still available in the other kha backends:
> https://github.com/Kode/Kha/blob/master/Backends/Kinc-hxcpp/kha/graphics4/VertexBuffer.hx
> https://github.com/Kode/Kha/blob/master/Backends/Krom/kha/graphics4/VertexBuffer.hx
> 
> Here are related Kha commits I found:
> [Kode/Kha@5c0bf12](https://github.com/Kode/Kha/commit/5c0bf12af2741259f56f321f9f2ba839bc960538)
> [Kode/Kha@2e456d4](https://github.com/Kode/Kha/commit/2e456d4dc7dcc750d63df110f126434f345f9083)
> 
> So the `lockInt16()` is being unified with `lock()`, but looks like it's not yet finished in the latest Kha.
> 
> (As a temporary workaround [armsdk/Kha](https://github.com/armory3d/Kha) will work in the meantime.)

sh-dave:
Well it's 12 commit ahead and 155 commits behind this repository. So i guess it's probably safer to use Lubos's fork as it might contain other changes necessary for iron/armory to work properly, as his Kha fork also seems to use forked submodules.

Lykdraft:
Aeh, the quote I posted was from Lubos himself, sooo.. I dunno... lol...

PR #1357, opened 2021-08-07
Fixes issues in KoreC Kinc-HL

QuantumCoderQC:
This PR fixes issues mentioned at https://github.com/armory3d/armory/issues/2267

Specifically, the following issues that occur in Visual Studio when targeting Kore-HL

```
'return Kore::Display::init()' : cannot convert from 'void' to 'int'
'Kore::Video::position': non-standard syntax; use '&' to create a pointer to member
'Kore::Video::finished': non-standard syntax; use '&' to create a pointer to member
'Kore::Video::duration': non-standard syntax; use '&' to create a pointer to member
'Kore::Mouse::unlock': function does not take 1 arguments
'Kore::Mouse::isLocked': function does not take 1 arguments
'Kore::Mouse::canLock': function does not take 1 arguments
```

Tested on Windows-HL and Android-HL targets with Armory3D projects and latest Kode repository.

### DIFF
--- a/Backends/Kinc-HL/KoreC/display.cpp
+++ b/Backends/Kinc-HL/KoreC/display.cpp
@@ -1,7 +1,7 @@
 #include <Kore/Display.h>
 #include <Kore/System.h>
 
-extern "C" int hl_kore_display_init() {
+extern "C" void hl_kore_display_init() {
 	return Kore::Display::init();
 }
 
--- a/Backends/Kinc-HL/KoreC/display.cpp
+++ b/Backends/Kinc-HL/KoreC/display.cpp
@@ -2,7 +2,7 @@
 #include <Kore/System.h>
 
 extern "C" void hl_kore_display_init() {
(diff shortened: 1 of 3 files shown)


Backends/Kinc-HL/KoreC/display.cpp
```diff
@@ -1,7 +1,7 @@
 #include <Kore/Display.h>
 #include <Kore/System.h>
 
-extern "C" int hl_kore_display_init() {
+extern "C" void hl_kore_display_init() {
 	return Kore::Display::init();
```
RobDangerous, line 5: This return needs to go.

RobDangerous (changes requested):
Commented on one little issue in the diff, everything else looks fine.

Backends/Kinc-HL/KoreC/display.cpp
QuantumCoderQC: Done

PR #1362, opened 2021-09-15
HL/C: correctly assign sample rate to sounds from .wav files

MoritzBrueckner:
If a sound was exported by Khamake for HL with quality 1.0, the sample rate wasn't correctly set because it was only set if `uncompress()` was called. However, sounds with quality 1.0 are already uncompressed so `uncompress()` directly returns, leaving the sample rate at 0. This PR fixes that.

@ RobDangerous: The length of the sound also isn't correctly set due to the same `hl.Ref` problem that recently occured for the global sample rate, however if I add the workaround from https://github.com/Kode/Kha/pull/1361 the length is still not correct and shows something like 5e-138. Either there is some issue with float conversion or the length isn't read correctly from the file. Should I open an issue for that?

### DIFF
--- a/Backends/Kinc-HL/KoreC/sound.cpp
+++ b/Backends/Kinc-HL/KoreC/sound.cpp
@@ -4,7 +4,7 @@
 #define STB_VORBIS_HEADER_ONLY
 #include <kinc/audio1/stb_vorbis.c>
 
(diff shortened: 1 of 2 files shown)


RobDangerous:
Yes yes, gimme all the issues, please.

RobDangerous:
Hu, wrong button.

PR #1370, opened 2021-10-11
renaming remaining texPosition's

sh-dave:
Looks like i missed even more places with the renames, and it turns out i searched for `"texPosition"` instead of only `texPosition`, so it's a brainbug instead of vscode :smile: 

There is also one i missed in `khamake/Data/psm/Texture.vcg`, is that still a thing?


RobDangerous:
We're fine, psm is not really a thing anymore.

PR #1377, opened 2021-11-17
add ssl support for hl/c builds

sh-dave:
- added hashlink's include/mbedtls and hashlink/zlib folders
- added hashlink's libs/fmt and libs/ssl folders
- added a copy of libs/fmt/fmt.c (named fmt_crypto.c)
  - stripped out the ogg/mp3/png/jpg code and only left the crypto code in


MoritzBrueckner:
Hi, I get some linker errors on Windows 10 and VS2019 after this PR:

```
1>ssl.obj : error LNK2001: Nicht aufgelöstes externes Symbol "__imp_CertOpenSystemStoreW".
1>ssl.obj : error LNK2001: Nicht aufgelöstes externes Symbol "__imp_CertEnumCertificatesInStore".
1>ssl.obj : error LNK2001: Nicht aufgelöstes externes Symbol "__imp_CertCloseStore".
```

"Nicht aufgelöstes externes Symbol" = "Unresolved external symbol"

sh-dave:
Looks like we might have to link against Crypt32.lib on windows for that https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certopensystemstorew

MoritzBrueckner:
Yes, that seems to work :)

sh-dave:
Cool thanks. Can you open a PR, i don't have access to a windows PC with all the necessary setup to test at the moment?

MoritzBrueckner:
Yes, do you think it belongs in the kincfile ([like this](https://github.com/Kode/Kha/blob/9ec02f81839bf83aab2867703575f3d1c3b33c3d/Backends/Kinc-HL/kincfile.js#L27), would be Kha only) or directly in Kincmake ([here](https://github.com/Kode/kincmake/blob/8371d2f016ca5bdd0ec643190b498b9be56b17e7/src/Exporters/VisualStudioExporter.ts#L608), but would also be included for hxcpp)?

sh-dave:
Hm, i think I'd add it to the kincfile, but let's summon @RobDangerous

RobDangerous:
It belongs in the kincfile.

PR #1393, opened 2022-01-19
HL: fix vertex data enum values

MoritzBrueckner:
New fix for https://github.com/Kode/Kha/pull/1384, this time with the big switch statement and only HL code modified. Tested with g2 and g4.


RobDangerous:
Well, this works but it should ideally be in C instead of Haxe so you can make use of Kinc's enum values directly - because that way we wouldn't need to change Kha if the values changed in Kinc.

Issue #872, opened 2018-09-05
Mouse wheel event doesn't trigger in IE


carlito767:
Hi,

The mouse wheel event doesn't trigger in Internet Explorer (version 11.228.17134.0).
Works fine for Chrome and Firefox.

PR #1417, opened 2022-04-24
Clean wheel event

RblSb:
https://caniuse.com/mdn-api_element_wheel_event
Basically the only working part of this code is `canvas.onwheel = mouseWheel`, but
> Internet Explorer only exposes the wheel event via addEventListener; there is no onwheel attribute on DOM objects

And `wheel mousewheel` is not a event at all, it seems.
Closes https://github.com/Kode/Kha/issues/872


RblSb:
Also `element.addEventListener("wheel")` works since Chrome 31, good thing that browserstack has one-minute trial for temp emails. Try better next time, mdn docs.

PR #1418, opened 2022-04-26
Minor webgl fixes

RblSb:
1. Do not push clear command at the end of frame if browser is not ie
2. Remove `bufferIndex + 1` to fully fill vertex buffers
3. If Vertex buffer is full, use `bufferData` instead of bufferSubData (should not have negative efffects, i hope)

More testing is better here, i just publish it to not to forget about

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -562,7 +562,7 @@ class SystemImpl {
 				#end
 
 				System.render([frame]);
-				if (SystemImpl.gl != null) {
+				if (ie && SystemImpl.gl != null) {
 					// Clear alpha for IE11
 					SystemImpl.gl.clearColor(1, 1, 1, 1);
 					SystemImpl.gl.colorMask(false, false, false, true);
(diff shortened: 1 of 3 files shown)


RblSb:
I think most of that `bufferIndex + 1` checks miss the opportunity to add last object to the buffer to fill it, if i dont miss some idea about it. Its only useful to fix for bunnymark stresstests anyway, so i'm not that curious.

RobDangerous:
We should really never use bufferData unless we are on Safari.

RblSb:
Yea, i guess that check is not related to gl buffer size to use `bufferData` anyway. Do you think that `if (ie && SystemImpl.gl != null)` change is ok or better to skip?

RobDangerous:
That's ok. With IE being less and less important it's good to make all IE fixes actual ie-only affairs.

RblSb:
Don't want to remove `+ 1` from size checks for now, too much responsibility. I hope you know where is squash button.

RobDangerous:
I'm a history-fan, I never squash.

PR #1427, opened 2022-08-24
fixing what seems to be a typo divide used instead of multiplying the…

onelsonic:
… scale devicePixelRatio to correctly resize the canvas.

Hi there, seems there is an error here :
if you want to apply the scale you should multiply it not divide, corrected it should be :
canvas.style.width = Std.int(clientWidth * scale) + "px";
canvas.style.height = Std.int(clientHeight * scale) + "px";

some info on devicePixelRatio can be found here:
https://webglfundamentals.org/webgl/lessons/webgl-resizing-the-canvas.html
current method is : 
  const displayWidth  = Math.round(canvas.clientWidth * dpr);
  const displayHeight = Math.round(canvas.clientHeight * dpr);

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -577,8 +577,8 @@ class SystemImpl {
 					canvas.width = clientWidth;
 					canvas.height = clientHeight;
 					if (scale != 1) {
-						canvas.style.width = Std.int(clientWidth / scale) + "px";
-						canvas.style.height = Std.int(clientHeight / scale) + "px";
+						canvas.style.width = Std.int(clientWidth * scale) + "px";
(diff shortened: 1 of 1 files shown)


RblSb:
You confused `canvas.width` and `canvas.style.width` in tutorial. Please test default Kha project without Armory to be sure that dpi stuff is correct here.
I think you probably want to check what css/resize logic is inside of Armory builds instead.

RobDangerous:
See https://github.com/Kode/Kha/commit/a0db16de6e05e8da213a7fcb5545174efb34543c#commitcomment-82027775

onelsonic:
ok so it seems like we have to hack our way around to have the correct width and height reflected across all devices display scales.

RobDangerous:
No hacking required when you use Kha as intended.

PR #1428, opened 2022-08-27
[HL] Fix type conversion for array values

MoritzBrueckner:
As discussed on the Discord server today. @ Rob, would sharing links to certain Discord messages get around the invitation system or is it safe to do?

I tried to test the changes as thoroughly as possible (overflows, etc.), but I couldn't test the implementation for big endian systems. Also, there seems to be an issue with rendering in the current HL backend (independent of my changes, I will open an issue soon), so I couldn't really test a larger project with my changes.


RobDangerous:
Looking good. I have no idea about Discord-links though.

PR #1440, opened 2022-12-03
Prevent formatting issue on file save

rpaladin:


RobDangerous:
Noo, Rob likes format on save.

PR #1445, opened 2022-12-21
fix vertical alignment for drawAlignedString

t3du:
vertical alignment is drawing: bottom, middle, and top with default as top. but it should be: top, middle, bottom, with default middle, just so it works as a text editor, and like horizontal alignment.


RobDangerous:
Sorry, that might maybe make more sense but it would break all code currently using that. But hey, it's time to kick out the whole thing anyway, deprecation-commit incoming...

PR #1463, opened 2023-06-12
expose kinc_get_mouse_position

bwkam:
never done this sort of thing before, so that's most likely wrong


RobDangerous:
Looks to me like this will do nothing. Did you test this at all?

RobDangerous:
Still looks like it will do nothing. And the HL code looks like it won't even compile. Please do at least a quick test before you send a pull request.

PR #1470, opened 2023-07-30
Fix Haxe 4.3 warnings

RblSb:


RobDangerous:
Hu but I don't get any warnings.

RblSb:
`@:enum` is deprecated on 4.3.1 (i'm using custom non-kha haxe version for now). Not sure about `@:extern`, both keywords are supported since 4.0, and haxe 3 is not supported anyway.

RobDangerous:
So it's about 4.3.1 and not just about 4.3 then. Oki, will merge this once I switch Kha to 4.3.1 as well.

PR #1471, opened 2023-09-10
G2 multitexturing

RblSb:
Don't know how to fix sampler warnings

### DIFF
--- a/Sources/Shaders/painter-image.frag.glsl
+++ b/Sources/Shaders/painter-image.frag.glsl
@@ -1,12 +1,37 @@
 #version 450
 
-uniform sampler2D tex;
+uniform sampler2D tex0;
+uniform sampler2D tex1;
+uniform sampler2D tex2;
+uniform sampler2D tex3;
+uniform sampler2D tex4;
+uniform sampler2D tex5;
+uniform sampler2D tex6;
+uniform sampler2D tex7;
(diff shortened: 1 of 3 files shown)


RobDangerous:
Sorry but I don't intend to change the G2 implementation like this. There are a thousand options to make 2D rendering faster for different situations and for different hardware. This for example will help in a lot of situations but it will also make the GPU side of things about ten times slower which is really bad for some mobile hardware.
Something like this is better kept as a sample-project than as a replacement for the standard G2 implementation.

RblSb:
Do you tested it on some project and gpu?
It shows big difference for my mobile gpus.
Since changes are minor, can we make selection between two shaders, with singletextured by default?

RobDangerous:
What's faster will depend entirely on what exactly your project is doing so just testing it on some project is pretty much pointless. 
We already have functionality for G2 selection - just copy it into your project and use that.

AndreiRudenko:
I agree that simpler G2 is better. You can use sprite sheets to have less texture switching.

RblSb:
You can provide your results from different gpus if you want to be helpful with one to ten texture switches, as realistic scenarios, or in real projects.
I'm still don't know how this can be 10 times slower in practice, since i only see no difference or 2-4x fps improvements in my low to high end mobile gpus.

PR #1475, opened 2023-10-10
Update ByteArray.hx

Onek8:
Fix hashlink warning for cast to Int pointed out by Zeta

On HL ByteArray.getUInt32 returns a hl.I64 for some reason: <https://github.com/Kode/Kha/blob/9585c2ba99683734dc2421ac31252c2a7e7e7153/Backends/Kinc-HL/kha/arrays/ByteArray.hx#L73>
But then it's turned into an Int here: <https://github.com/Kode/Kha/blob/9585c2ba99683734dc2421ac31252c2a7e7e7153/Sources/kha/arrays/Uint32Array.hx#L16>, which leads to that new warning.


RobDangerous:
But this is wrong, this should be fixed in HL.

PR #1477, opened 2023-11-04
Type Dynamic callbacks in Assets

RblSb:
I spended enough hours today debugging why this (not my) silly code breaks on cpp:
```haxe
Assets.loadEverything(function () {
	// Avoid passing update/render directly,
	// so replacing them via code injection works
	Scheduler.addTimeTask(function () { update(); }, 0, 1 / 60);
	System.notifyOnFrames(function (frames) { render(frames); });
}, (err:AssetError) -> {
	throw err;
});
```
I want to improve this api, which will also fix future dynamic problems, like `dynamicObj.name.contains` failure on some static targets.
With typed callbacks this should be harder to break and easier to use now.
Also, is this possible to get shaders in assets loader? Seems like not, or there is some ways to enable runtime loading?

### DIFF
--- a/Sources/kha/Assets.hx
+++ b/Sources/kha/Assets.hx
@@ -5,27 +5,31 @@ import haxe.Unserializer;
 
 using StringTools;
 
-typedef AssetData = {
-	name: String,
-	files: Array<String>,
-	file_sizes: Array<Int>,
-	type: String,
-	// image
-	?original_width: Int,
-	?original_height: Int,
-	// shader
-	?inputs: Array<AssetShaderVar>,
-	?outputs: Array<AssetShaderVar>,
-	?uniforms: Array<AssetShaderVar>,
-	?types: Array<{
-		name: String,
-		members: Array<AssetShaderVar>,
-	}>
+private typedef AssetDataObject = {
+	/** File name, given by khamake, used as identifier in `Assets.someList.get()` function **/
+	var name: String;
+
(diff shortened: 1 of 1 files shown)


RobDangerous:
I'm sorry but... I don't know what you are talking about.

RblSb:
This PR solves two problems:
```haxe
Assets.loadEverything(() -> {}, (err:AssetError) -> {
	throw err;
});
```
Will be detected as `(callback, null, null, failed)`, instead of `(callback, filter)`, since callbacks are not Dynamic anymore.
And it also solves problem, when you try to access `item.name` fields, like:
```haxe
Assets.loadEverything(() -> {}, item -> {
return item.name.contains("menu");
});
```
Since `item.name` will be typed as `String`, and not `Dynamic` with unknown native `contains` method (for example, js has `str.includes`, but no `str.contains`)

RobDangerous:
OK. But I originally made those dynamic because I wanted to add the option to add any data to assets in the khafile. Can typedefs support that?

RblSb:
There is no `typedef A = {foo:Int} & Dynamic;` support in haxe yet, but if you think about making this feature before that support, i can change this typedef to `abstract`, that will provide completion for `item.name` / other fields with good types, but also allow `item.anythingElse` fields and return Dynamic values from them.

RobDangerous:
Yes, let's make some abstract art.

PR #1487, opened 2024-01-31
fix audio callback init.

e2002e:
There is a missing argument in kinc_a2_set_callback,
Adding null I was able to compile the project.


RobDangerous:
Yes, the GitHub Actions already told me about it. But unfortunately I can't merge this because you changed the Tools submodules.

PR #1488, opened 2024-01-31
fix audio callback init

e2002e:
I'm opening another pull request from my main modified branch.
I rolled back the tools.


RobDangerous:
Jippie!

PR #1490, opened 2024-03-14
Bypass browser cache on reloading image with the same name.

AndreiRudenko:
Not sure if it's good, the intention was to have reloading changed images with the same name(path) working on electron. Could this be bad for browser games in terms of loading time?


RobDangerous:
Yes, looks like this might be bad for loading times. But you could make it an electron only thing by using the kha_debug_html5 define.

PR #1504, opened 2025-03-23
Don't throw on small determinant

RblSb:
I think this is too easy to get this error while inverting matrices with a small scale and such check can be done externally when needed. Is there are cases when this throw can be useful?


RobDangerous:
This does not look good to me. Sorry, I'd rather have it error instead of returning a silly matrix.

PR #1505, opened 2025-04-27
Add stereo panes for Mobile HTML5

moisesjpelaez:
Add stereo panes to `MobileWebAudioChannel` to make 2D/3D audio work on Mobile browsers. Tested with the [Aura](https://github.com/MoritzBrueckner/aura) audio library.

### DIFF
--- a/Backends/HTML5/kha/js/MobileWebAudioChannel.hx
+++ b/Backends/HTML5/kha/js/MobileWebAudioChannel.hx
@@ -18,7 +18,7 @@ class MobileWebAudioChannel implements kha.audio1.AudioChannel {
 
 	var leftGain: GainNode;
 	var rightGain: GainNode;
-	var stereoGain: GainNode;
+	var attenuationGain: GainNode;
 	var splitter: ChannelSplitterNode;
 	var merger: ChannelMergerNode;
 
@@ -40,7 +40,7 @@ class MobileWebAudioChannel implements kha.audio1.AudioChannel {
 		leftGain = MobileWebAudio._context.createGain();
 		rightGain = MobileWebAudio._context.createGain();
 		merger = MobileWebAudio._context.createChannelMerger(2);
-		stereoGain = MobileWebAudio._context.createGain();
+		attenuationGain = MobileWebAudio._context.createGain();
 		gain = MobileWebAudio._context.createGain();
 
 		source.connect(splitter);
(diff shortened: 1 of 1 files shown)


RobDangerous:
Hi!
Uhm, I'm not sure what any of that means. How is that used? Why does it need to be in Kha instead of just putting it in Aura?

moisesjpelaez:
Hello!

Aura is an audio library for Armory3D and Kha (from what I've seen, I think it relies on Kha for that). I saw its internal [Html5StreamChannel](https://github.com/MoritzBrueckner/aura/blob/47effabbc44065ade0255ad5f74a955b481bf782/Sources/aura/channels/Html5StreamChannel.hx) gets access to the `volume` property from `MobileWebAudioChannel` to set the sound volume on mobile browsers. 

From Aura's `HTML5StreamChannel`:
`override function parseMessage(message: Message) {
		switch (message.id) {
			// Because we're using a HTML implementation here, we cannot use the
			// LinearInterpolator parameters
			case ChannelMessageID.PVolume: khaChannel.volume = cast message.data;
			case ChannelMessageID.PPitch: @:privateAccess khaChannel.source.playbackRate.value = dopplerRatio * cast message.data;
			case ChannelMessageID.PDopplerRatio: dopplerRatio = cast message.data;
			case ChannelMessageID.PDstAttenuation: @:privateAccess khaChannel.stereoGain.gain.value = cast message.data;
			case ChannelMessageID.PVolumeLeft: @:privateAccess khaChannel.leftGain.gain.value = cast message.data;
			case ChannelMessageID.PVolumeRight: @:privateAccess khaChannel.rightGain.gain.value = cast message.data;

			default:
				super.parseMessage(message);
		}
	}`

The Pitch and the rest of the rest of the cases in `parseMessage` were empty and I realized by accessing the `source.playbackRate.value` I could assign the pitch. For 3D audio support each speaker needed independent volume and that's what I added in this PR, with the `leftGain` and `rightGain` the volume can be assigned to each left and right speaker independently.

I added an extra `stereoGain` to control the attenuation independently from the volume, and the `gain` property keeps controlling the overall volume.

RobDangerous:
But I don't see why this needs to be in Kha and this makes things slower for everything else. Why not just do all of that from the outside after you created a channel? You can even add the properties to the object without subclassing the channel because it gets compile to JavaScript.

moisesjpelaez:
Took me a while to understand but I got it. Thanks for pointing it out! Actually made and tested the changes in Aura.

RobDangerous:
Nice 👍

PR #1530, opened 2026-04-11
Small updates

moisesjpelaez:
- ~put `matrixCache` static variables outside `setMatrix` functions for 3D performance improvements in WebGL (taken from Armory's Kha)~
- remove deprecation warnings when converting to a UInt32 in Kore-HL
- expose `setMouseCursor` in Krom


RblSb:
I think moving `matrixCache` static var outside of function only makes it available inside of class and makes generated name without `setMatrix3_` prefix?

moisesjpelaez:
That's correct, sorry, I reverted the change.

PR #1532, opened 2026-04-28
Assets.images previews in field documentation

RblSb:
Adds image link to field doc, so you can see / go to original image, it looks like this:

image


RobDangerous:
Very nice!

PR #1533, opened 2026-05-02
Faster audio decompression on html5

RblSb:
Makes background audio decoding 30x faster (almost) without frame drops.
4 minutes music is decompressed in 0.6s instead of 18 seconds. On my androids it drops to 50-55 fps at the end of decompression for long tracks, but since most decompressions happen only for sounds and at `loadEverything` call, i think this trade-off is worth it.

Reference info about setTimeout slowness:
https://nolanlawson.com/2025/08/31/why-do-browsers-throttle-javascript-timers/


RobDangerous:
Funny. Thanks.

PR #1534, opened 2026-05-04
RGBA64U texture format

RblSb:
Also global `UNPACK_PREMULTIPLY_ALPHA_WEBGL` activation is removed, because it's only needed when loading DOM images and breaks image generation from bytes


RobDangerous:
Kore 2 doesn't currently have that. Do you plan to add it? Don't want to add a texture format that's only supported for html5.

PR #1537, opened 2026-07-26
Implement keyboard.show for html5

### LINKED ISSUE #1257: Handle .show()/.hide() of the virtual keyboard on mobile devices with HTML5 target
**Is your feature request related to a problem? Please describe.**
Currently on mobile OS using the HTML5 target in a browser (tested on Android with Chrome and Firefox), the call to show the virtual keyboard (`Keyboard.get().show()`) gets ignored and thus makes it impossible to use text input on such devices.

**Describe the solution you'd like**
I would like the HTML5 target to distinguish whether it is running inside a browser on a touch only device with no physical keyboard. If that is the case, calls to `Keyboard.get().show() / .hide()` should show/hide the virtual keyboard and handle text input properly.

**Describe alternatives you've considered**
I've considered to also export as native apps to iOS and Android and make my website detect when it is running on those platforms and displaying links to the respective store. This solution takes up additional time (publishing on the stores and passing app screening), comes with added cost (developer accounts on both platforms) and potentially deters some users because they first have to download an app.
Another solution for my particular case, where the input is only numbers based is to create a *number pad* widget in my app that gets displayed when a mobile browser is detected. This solution is only viable for number only input and doesn't address the underlying problem.

**Additional context**
I found something about the topic on SO [show-virtual-keyboard-on-mobile-phones-in-javascript](https://stackoverflow.com/questions/6837543/show-virtual-keyboard-on-mobile-phones-in-javascript) that could potentially be helpful to solve this. IIRC in SDL2 one can specify the rect where the text input takes place, if something like that would be added to Kha, maybe it would be possible for Kha to add an invisible layer on top of text inputs, that triggers the virtual keyboard? Not sure, not really an area of expertise of mine.

RblSb:
`keyboard.show()` now creates hidden focused input html element to show virtual keyboard, resends keydown/up events to game canvas and generates direct keypress events when input value updated.

Closes https://github.com/Kode/Kha/issues/1257

### DIFF
--- a/Backends/HTML5/kha/SystemImpl.hx
+++ b/Backends/HTML5/kha/SystemImpl.hx
@@ -20,7 +20,9 @@ import kha.graphics4.TextureFormat;
 import kha.input.Gamepad;
 import kha.input.KeyCode;
 import kha.input.Keyboard;
+import kha.input.KeyboardImpl;
 import kha.input.Mouse;
+import kha.input.MouseImpl;
 import kha.input.Sensor;
 import kha.input.Surface;
 import kha.js.AudioElementAudio;
@@ -215,7 +217,8 @@ class SystemImpl {
 	}
 
 	public static function vibrate(ms: Int): Void {
-		if (Browser.navigator.vibrate == null) return;
+		if (Browser.navigator.vibrate == null)
+			return;
 		Browser.navigator.vibrate(ms);
 	}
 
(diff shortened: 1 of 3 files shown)


RblSb:
@RobDangerous i just found another reason why this PR is nice, it fixes keypress events with disabled system interventions:
```haxe
Keyboard.disableSystemInterventions(Full);
Keyboard.get().notify(null, null, char -> trace(char)); // not working right now
```
For example, i would like to prevent default `Space` key from scrolling page in some browsers, but then you cannot type spaces into canvas text inputs.

I did too much verbose changes as always, so i can cleanup it to make diff more readable for review if you want.

RobDangerous:
That would be nice, cleanups can go into a separate PR.

RblSb:
Since there is no way to prevent page scroll on `Space` / some other shortcuts like `/`, without disabling `keypress` at all, i decided to use `keydown` for both events. But i just found that `Ctrl-C` will generate `c` char now, so i also added `isPrintableText`, that will check for modifiers and don't generate chars when meta/ctrl/alt is active, except when `AltGr` chars are typed.

`KeyboardImpl` is still just:
> keyboard.show() now creates hidden focused input html element to show virtual keyboard, resends keydown/up events to game canvas and generates direct keypress events when input value updated.

And `onBlur`/`onFocus` ifs are needed to prevent `System.background/foreground` callbacks when this input element will change focus on `keyboard.show/hide`.

